Issue #3, opened 2026-05-26
Tracking: Lifecycle Manager + Session Manager lane


harshitsinghbhandari:
Tracking issue for the **Lifecycle Manager (LCM) + Session Manager (SM)** lane of the backend rewrite (Go daemon).

## Goal

A deterministic supervisor for agent sessions on the **OBSERVE → DECIDE → ACT** loop:
- **LCM** — a synchronous reducer. Observers push raw facts; it runs a *pure* decision, persists, and fires reactions. Never polls.
- **SM** — the explicit-mutation plumbing: spawn / kill / restore / cleanup. Reports outcomes to the LCM; never derives state itself.

## Approach: contract-first, single integration branch

`feat/lcm-sm-contracts` is the **lane integration branch**: every sub-PR branches off it and merges **into** it; the whole lane lands on `main` as one unit once complete.

## PR sequence — implementation COMPLETE ✅

- [x] **Repo skeleton** — `backend/` + `frontend/` — #1 *(merged to main)*
- [x] **Contracts** — `domain/` + `ports/` + CI workflows — #2 *(= the integration branch)*
- [x] **decide core** — pure deciders + truth-table tests — #4
- [x] **LCM — pipeline** *(split A)* — `Apply*` pipeline, per-session serialisation, store integration — #5
- [x] **LCM — reactions** *(split B)* — reaction table + escalation engine + real `TickEscalations` — #6
- [x] **Session Manager** — spawn / kill / restore / cleanup / list, against fakes — #7
- [ ] **Integration** — swap fakes for real adapters as the other lanes land *(incremental, next phase)*
- [ ] **Land lane → main** — review `feat/lcm-sm-contracts` as the complete lane and merge to `main`.

All implemented behind fakes; `gofmt`/`build`/`vet`/`test -race` green across domain, decide, lifecycle, session.

## Integration phase — carried-forward items (must address as real adapters land)

- **`react()` out-of-lock dispatch** (#6) — reactions fire after the per-session lock releases; under concurrent observers a dispatch can run on a stale snapshot / out of order. Documented in `reactions.go`; needs a per-session react queue or a re-check before dispatch once the live daemon runs concurrent observers.
- **`ExpectedRevision` unused** (#5) — in-process per-session mutex covers a single daemon; multi-writer/CDC would need the optimistic-CAS the contract already exposes.
[...]

PR #4, opened 2026-05-26
feat(decide): pure DECIDE core + exhaustive truth-table tests

harshitsinghbhandari:
## Summary
Implements the pure DECIDE core for the LCM + Session Manager lane (PR #3), replacing the stubbed `panic(...)` bodies in `backend/internal/domain/decide/decide.go` with real, total, side-effect-free deciders, plus exhaustive table tests.

> **Stacked on #2** (`feat/lcm-sm-contracts`) — this PR is based on the contracts branch, not `main`, so the diff is just the two decide files. Merge #2 first.

### Deciders
- **`ResolveProbeDecision`** — kill-intent short-circuits straight to terminal `killed`; a *failed* probe (timeout/error) and probe disagreement (runtime/process mismatch, or runtime-dead with recent activity) route to `detecting`; **only** runtime-dead + process-dead + no-recent-activity concludes `killed`. (Honors "failed probe ≠ dead" and "one authority for death".)
- **`ResolveOpenPRDecision`** — the PR pipeline ladder in exact order: `ci_failing → changes_requested → approved+mergeable → approved → review_pending → idle-beyond → open`.
- **`ResolveTerminalPRStateDecision`** — merged → idle / `merged_waiting_decision`; closed → idle.
- **`CreateDetectingDecision`** — anti-flap quarantine. Unchanged-evidence counter; `StartedAt` is preserved across the whole detecting episode so the duration cap is a real wall-clock safety net even when evidence flaps. Escalates to `stuck` at `DetectingMaxAttempts` (3) consecutive unchanged ticks **or** `DetectingMaxDuration` (5m).
- **`HashEvidence`** — strips ISO datetimes, bare time-of-day, and epoch numbers, then collapses whitespace before SHA-256, so restamped-but-unchanged signals compare equal.

### Notes for reviewers
- No changes to the frozen contract types in `domain/` or `ports/`; no `decide` input struct needed extending.
- `LifecycleDecision` carries no runtime sub-state fields, so the probe decider expresses runtime death via `SessionReason` (`runtime_lost` / `agent_process_exited`) — the LCM is expected to translate that into the persisted runtime substate.
- `DetectingInput.ProposedState` is kept for the contract but isn't load-bearing in the output (output state is always `detecting` or escalated `stuck`); the reason is driven by `ProposedReason`.

## Test plan
- [x] `gofmt -l .` prints nothing
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -race ./...`
- [x] 100% statement coverage on `internal/domain/decide`
- [x] Every branch of every decider table-tested, plus a consistency test that the open-PR ladder's display `Status` matches `DeriveLegacyStatus` over the canonical state it emits, and an end-to-end probe→detecting→stuck escalation flow

### DIFF
--- a/backend/internal/domain/decide/decide.go
+++ b/backend/internal/domain/decide/decide.go
@@ -166,7 +166,11 @@ func ResolveOpenPRDecision(in OpenPRInput) LifecycleDecision {
 		return base(domain.StatusCIFailed, domain.PRReasonCIFailing, domain.SessionWorking, domain.ReasonFixingCI)
 	case in.ChangesRequested:
 		return base(domain.StatusChangesRequested, domain.PRReasonChangesRequested, domain.SessionWorking, domain.ReasonResolvingReviewComments)
-	case in.Approved && in.Mergeable:
+	case in.Mergeable:
+		// Mergeability is the authoritative merge gate, so it already folds in
+		// "approved if review is required". Checking it before Approved means a
+		// PR on a no-required-review repo (mergeable, not formally approved) is
+		// still surfaced as ready-to-merge instead of falling through to PR_OPEN.
 		return base(domain.StatusMergeable, domain.PRReasonMergeReady, domain.SessionIdle, domain.ReasonAwaitingExternalReview)
 	case in.Approved:
 		return base(domain.StatusApproved, domain.PRReasonApproved, domain.SessionIdle, domain.ReasonAwaitingExternalReview)
@@ -270,6 +274,11 @@ func HashEvidence(evidence string) string {
 // timestampPatterns strip the time-varying parts of an evidence string before
 // hashing. Order matters: the full datetime form is removed before the bare
 // time-of-day and epoch forms so they don't partially match.
+//
+// The epoch pattern strips any bare 10-13 digit run (unix seconds/millis). That
+// is broad enough to also clobber a same-width numeric ID if one ever appears in
+// an evidence string; evidence is decider-authored, so today nothing else lands
+// in that range, but keep IDs out of evidence strings to preserve hash fidelity.
 var timestampPatterns = []*regexp.Regexp{
 	regexp.MustCompile(`\d{4}-\d{2}-\d{2}[T ]\d{2}:\d{2}:\d{2}(?:\.\d+)?(?:Z|[+-]\d{2}:?\d{2})?`),
 	regexp.MustCompile(`\d{2}:\d{2}:\d{2}(?:\.\d+)?`),
--- a/backend/internal/domain/decide/decide.go
+++ b/backend/internal/domain/decide/decide.go
@@ -212,14 +212,22 @@ func HashEvidence(evidence string) string {
 	return hex.EncodeToString(sum[:])
 }
 
-// timestampPatterns strip the time-varying parts of an evidence string before
-// hashing. Order matters: the full datetime form is removed before the bare
-// time-of-day and epoch forms so they don't partially match.
+// timestampPatterns is the list of regexes HashEvidence applies (in order) to
+// delete the time-varying parts of an evidence string before hashing, so the
+// same ambiguous signal restamped with a new clock value hashes equal and the
+// detecting counter keeps climbing instead of resetting every tick.
 //
-// The epoch pattern strips any bare 10-13 digit run (unix seconds/millis). That
-// is broad enough to also clobber a same-width numeric ID if one ever appears in
-// an evidence string; evidence is decider-authored, so today nothing else lands
-// in that range, but keep IDs out of evidence strings to preserve hash fidelity.
+// Order matters: the full datetime form is removed first so its embedded
+// HH:MM:SS isn't half-eaten by the bare time-of-day pattern that follows.
+//
+//   1. full ISO-8601 / RFC3339 datetime — date, a T or space separator,
+//      HH:MM:SS, optional fractional seconds, optional Z or ±HH:MM offset.
+//      e.g. "2026-05-26T12:00:00Z", "2026-05-26 12:00:00.218+05:30"
+//   2. a bare time-of-day, e.g. "12:00:00" or "12:00:00.218"
+//   3. a bare unix epoch — any 10-13 digit run (seconds or millis), e.g.
+//      "1716724800". This is broad enough to also clobber a same-width numeric
+//      ID if one ever appears in evidence; evidence is decider-authored, so keep
+//      IDs out of evidence strings to preserve hash fidelity.
 var timestampPatterns = []*regexp.Regexp{
 	regexp.MustCompile(`\d{4}-\d{2}-\d{2}[T ]\d{2}:\d{2}:\d{2}(?:\.\d+)?(?:Z|[+-]\d{2}:?\d{2})?`),
 	regexp.MustCompile(`\d{2}:\d{2}:\d{2}(?:\.\d+)?`),
--- a/backend/internal/domain/decide/decide.go
+++ b/backend/internal/domain/decide/decide.go
@@ -2,13 +2,14 @@
 // collapses observed facts (plus the prior detecting/activity memory) into one
 // LifecycleDecision. Every function here must remain side-effect free so the
 // whole status truth-table can be tested in isolation.
-//
-// NOTE: function bodies are stubbed in this contracts PR. The real logic + the
-// exhaustive truth-table tests land in the follow-up "decide core" PR. The
-// signatures and the input/output shapes are what we are stabilising now.
 package decide
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"regexp"
+	"strings"
 	"time"
 
 	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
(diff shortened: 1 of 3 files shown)


harshitsinghbhandari (commented):
## Review — pure DECIDE core

**Recommendation: Approve, with follow-ups.** Verified independently in a throwaway worktree: gofmt clean, `go build`/`vet` pass, `go test -race` green at **100% statement coverage**. The core logic is correct, pure, and exhaustively tested. Findings below are about the seam with the (not-yet-built) LCM, one semantic gap, and minor items — none are correctness bugs in the isolated functions.

### The two areas called out in the request — both correct
- **Probe branch ordering** ✓ `kill → any-probe-failed → runtime switch`. A failed probe never concludes death; disagreements and indeterminate-process quarantine; only `runtime-dead + process-dead + no-recent-activity` reaches `killed`.
- **Detecting counter/duration** ✓ Preserving `StartedAt` across an evidence change while resetting the attempt counter is the right call — the 5-min cap is a whole-episode wall-clock safety net, not per-evidence, so a flapping-but-stuck session still escalates. The "duration cap fires even when evidence keeps flapping" test proves it.

### Architecture & design (carry into the LCM PR)
1. **Probe decider conflates liveness with session activity state.** A healthy runtime returns `SessionWorking / ReasonTaskInProgress`. When the reaper feeds this through `ApplyRuntimeObservation`, the LCM must not let a liveness verdict clobber a more specific activity-derived state (e.g. a session in `waiting_input` set by `ApplyActivitySignal`). Needs an explicit composition rule for *which decider owns `SessionState` when the runtime is alive*. Not a bug here — a contract to document in the LCM PR.
2. **`mergeable` is unreachable without `Approved`.** The distilled ladder is "approved **/none** + mergeable → MERGEABLE", but the impl only reaches `MERGEABLE` on `Approved && Mergeable`. A mergeable PR on a repo with no required review falls through to `PR_OPEN`. Either `OpenPRInput` should express "no review required" or `Mergeable` should be weighed independently. Confirm intent before the LCM builds the `SCMFacts → OpenPRInput` mapping.

### Integration contract (LCM PR)
3. **Detecting-memory lifecycle.** The deciders correctly return `Detecting == nil` for every non-detecting outcome and non-nil only while quarantined. The LCM must therefore clear the persisted detecting memory (via `LifecyclePatch.ClearDetecting`) whenever a decision's `Detecting` is nil — otherwise a stale `StartedAt`/`EvidenceHash` leaks into a later episode.

### Testing
4. **Extend the consistency cross-check.** `TestResolveOpenPRDecisionDerivesConsistently` (`decision.Status == DeriveLegacyStatus(canonical)`) is excellent but only covers the open-PR ladder. The probe and terminal deciders are also consistent (hand-verified) — adding them to the same test locks the invariant against drift.

### Code quality (minor)
5. **`HashEvidence` epoch regex `\b\d{10,13}\b`** strips any 10–13 digit run, which would also clobber a legitimate large ID if one ever appears in an evidence string. Low risk today; worth a comment or tightening. (`HH:MM` without seconds also isn't stripped — fine for the current full-timestamp format.)

Nice work — the truth-table is thorough and the anti-flap semantics are right.

harshitsinghbhandari:
Thanks for the thorough review, @harshitsinghbhandari. Addressed in 9920c6d:

**#2 — mergeable unreachable without `Approved`** ✅ Fixed. `ResolveOpenPRDecision` now keys `MERGEABLE` on `Mergeable` alone, checked *before* `Approved`. Since `Mergeability` is the authoritative merge gate (it already folds in approval where review is required), a mergeable PR on a no-required-review repo is now surfaced as ready-to-merge instead of falling through to `PR_OPEN`. The two explicitly-specified cases are preserved: `approved && mergeable → MERGEABLE` (caught by the mergeable branch first) and `approved && !mergeable → APPROVED`. Added a `mergeable && !approved → MERGEABLE` test case.

**#4 — extend the consistency cross-check** ✅ Done. Renamed to `TestDecidersDeriveConsistently`; it now asserts `decision.Status == DeriveLegacyStatus(canonical)` across the open-PR ladder **and** the probe + terminal deciders. (The terminal `none`/`open` default is excluded — it's a documented no-op for misuse, and its `PR_OPEN`/`working` shape is intentionally not derive-consistent.)

**#5 — `HashEvidence` epoch regex breadth** ✅ Documented. Added a comment noting `\b\d{10,13}\b` will also strip a same-width numeric ID, and that evidence strings should keep IDs out to preserve hash fidelity. Left the pattern as-is since evidence is decider-authored.

**#1 — probe decider conflates liveness with activity state** — Agreed, deferring to the LCM PR. This is a composition-rule question (which decider owns `SessionState` when the runtime is alive), not a bug in the pure function. Will document an explicit precedence rule so a liveness `working` verdict can't clobber a more specific activity-derived state (e.g. `waiting_input`) in `ApplyRuntimeObservation`.

**#3 — detecting-memory lifecycle** — Agreed, LCM-PR contract. The deciders already return `Detecting == nil` for every non-detecting outcome; the LCM must clear persisted detecting memory (`LifecyclePatch.ClearDetecting`) whenever a decision's `Detecting` is nil so a stale `StartedAt`/`EvidenceHash` can't leak into a later episode. Will wire this when the LCM lands.

Gates still green: gofmt clean, build/vet/`test -race`, 100% statement coverage.

harshitsinghbhandari (commented):
## Re-review — 9920c6d

Verified independently (throwaway worktree on 9920c6d): gofmt clean, `build`/`vet` pass, `go test -race` green at **100% coverage**. All actionable findings resolved:

- **#2 mergeable-without-approval** — correct. Moving `Mergeable` ahead of `Approved` but still behind `CIFailing`/`ChangesRequested` is a genuine strict superset: every previously-`MERGEABLE` case is unchanged, the no-required-review case is now surfaced, and CI-fail / changes-requested precedence is preserved. The "authoritative merge gate" comment captures the rationale.
- **#4 consistency cross-check** — now spans the probe + terminal deciders (`TestDecidersDeriveConsistently`), with the terminal none/open no-op reasonably excluded. This locks decider output against `DeriveLegacyStatus` drift.
- **#5 epoch-regex breadth** — documented.

**#1** (liveness-vs-activity composition) and **#3** (detecting-memory clearing) acknowledged as LCM-PR contract follow-ups — agreed, those belong there.

LGTM. Good, fast turnaround.

backend/internal/domain/decide/decide.go
harshitsinghbhandari: Why are types defined here, can we not define types in execution files

backend/internal/domain/decide/decide.go
```diff
@@ -82,27 +83,217 @@ type DetectingInput struct {
...
+// The epoch pattern strips any bare 10-13 digit run (unix seconds/millis). That
+// is broad enough to also clobber a same-width numeric ID if one ever appears in
+// an evidence string; evidence is decider-authored, so today nothing else lands
+// in that range, but keep IDs out of evidence strings to preserve hash fidelity.
+var timestampPatterns = []*regexp.Regexp{
+	regexp.MustCompile(`\d{4}-\d{2}-\d{2}[T ]\d{2}:\d{2}:\d{2}(?:\.\d+)?(?:Z|[+-]\d{2}:?\d{2})?`),
+	regexp.MustCompile(`\d{2}:\d{2}:\d{2}(?:\.\d+)?`),
+	regexp.MustCompile(`\b\d{10,13}\b`),
+}
+
```
harshitsinghbhandari, line 249: What exactly is this doing?

backend/internal/domain/decide/decide.go
harshitsinghbhandari: Good call — moved the decider input/output type definitions (LifecycleDecision, ProbeInput, ProcessLiveness + its consts, OpenPRInput, DetectingInput) into a dedicated decide/types.go in 4d8a206. decide.go now holds only the package doc, the anti-flap tuning consts, and the decider functions. Resolving.

backend/internal/domain/decide/decide.go
harshitsinghbhandari: That is the unexported detecting() helper: a thin adapter that funnels a probe verdict into the shared anti-flap path. Each probe branch that lands on ambiguity calls it with a proposed reason + evidence string; it packages those (plus the prior counter from the same ProbeInput) into a DetectingInput and defers to CreateDetectingDecision, so probe-driven ambiguity is counted/escalated by the exact same quarantine logic instead of every branch re-implementing the counter. Expanded its doc comment to say this explicitly in 4d8a206. Resolving.

harshitsinghbhandari: Correction — my previous reply answered the wrong element; you meant the `timestampPatterns` regex block. Here's what it actually does: `HashEvidence` applies these three regexes in order to delete the time-varying parts of an evidence string *before* hashing, so the same ambiguous signal restamped with a new clock value hashes equal and the detecting counter keeps climbing instead of resetting every tick.

1. `\d{4}-\d{2}-\d{2}[T ]\d{2}:\d{2}:\d{2}(?:\.\d+)?(?:Z|[+-]\d{2}:?\d{2})?` — a full ISO-8601/RFC3339 datetime: date, a `T` or space, `HH:MM:SS`, optional fractional seconds, optional `Z` or `±HH:MM` offset. Matches e.g. `2026-05-26T12:00:00Z` and `2026-05-26 12:00:00.218+05:30`.
2. `\d{2}:\d{2}:\d{2}(?:\.\d+)?` — a bare time-of-day like `12:00:00` / `12:00:00.218`.
3. `\b\d{10,13}\b` — a bare unix epoch (10-13 digit run, seconds or millis), e.g. `1716724800`.

Order matters: the full datetime is stripped first so its embedded `HH:MM:SS` isn't half-eaten by pattern 2. Documented all of this inline in fbfbcd5 (+ gofmt in cdfcdd2).

backend/internal/domain/decide/decide.go
```diff
@@ -22,87 +23,228 @@ const (
...
-	panic("decide.ResolveOpenPRDecision: not implemented (decide-core PR)")
+	base := func(status domain.SessionStatus, prReason domain.PRReason, ss domain.SessionState, sr domain.SessionReason) LifecycleDecision {
+		return LifecycleDecision{
+			Status:        status,
+			SessionState:  ss,
+			SessionReason: sr,
+			PRState:       domain.PROpen,
+			PRReason:      prReason,
+		}
+	}
```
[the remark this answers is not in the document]
harshitsinghbhandari: Fixed in 918c5b4. ResolveOpenPRDecision now sets a stable, timestamp-free Evidence summary for every ladder outcome — `<condition> #<num> <url>`, e.g. `ci_failing #123 https://...` or `merge_ready #7` — which also consumes the previously-unused OpenPRInput.Number/URL identity inputs. Added TestResolveOpenPRDecisionEvidence to lock the format (number-only and no-identity fallbacks included). Resolving.

backend/internal/domain/decide/types.go
```diff
@@ -0,0 +1,66 @@
...
+// and the (possibly updated) detecting memory.
+type LifecycleDecision struct {
+	Status        domain.SessionStatus
+	Evidence      string
+	Detecting     *domain.DetectingState
+	SessionState  domain.SessionState
+	SessionReason domain.SessionReason
+	PRState       domain.PRState
+	PRReason      domain.PRReason
+}
```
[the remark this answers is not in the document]
harshitsinghbhandari: Good catch on the seam risk. Of your two options, always setting PRNone is unsafe — a probe/detecting/kill verdict knows nothing about the PR, so writing PRNone on a routine probe tick would clobber a live PR. The other concern (representable unchanged) is already handled: an empty PRState (`""`) is distinguishable from every valid PRState (none/open/merged/closed), so it is a clean "not addressed" sentinel, and LifecyclePatch.PR is *already* `*domain.PRSubstate` with documented nil-means-unchanged semantics.

Rather than add enum-pointer boilerplate (Go can't take the address of a const, and pointer enums would be inconsistent with the value-enum style used throughout domain/), I documented the contract explicitly on LifecycleDecision in 918c5b4: zero-value sub-state means "leave unchanged", and the LCM must map an empty PRState to a nil patch.PR rather than writing PRNone. This mirrors the existing Detecting/ClearDetecting three-way semantics. Resolving.

PR #16, opened 2026-05-27
fix: handle SCM observer seam facts

harshitsinghbhandari:
## Summary

Fixes the three SCM observer seam issues called out in /Users/harshitsinghbhandari/.me/notes/ao-scm-observer-review.md:

1. Add explicit SCMFacts.Draft handling so state=open,draft=true is treated as a draft PR, with CI failure still dominating drafts.
2. Wire PendingComments into the open-PR decision path so bot comments drive the bugbot-comments reaction while human comments continue to route to changes-requested.
3. Consume merge-conflict mergeability facts into a distinct PR reason so the merge-conflicts reaction can fire.

## Validation

From backend/:
- go build ./...
- go vet ./...
- go test -race ./...


harshitsinghbhandari (commented):
LGTM

Issue #22, opened 2026-05-28
feat(api): 1b·c — Issues, PRs, runtime & remaining route shell


neversettle17-101:
Part of #10 — Go HTTP Daemon lane. Sub-issue of umbrella #18.

> ⚠️ **Scope:** Placeholder. Register the remaining ~12 routes as `501 Not Implemented` with structured `planned` bodies. Final route count + REST audit pending a TS-source read pass; opening this issue now to track scope.

## Depends on
- #20 (#18a Projects) and #21 (#18b Sessions) for shared scaffolding.

## Tentative route list (subject to TS audit)

- `GET /api/v1/issues` — list issues
- `GET /api/v1/backlog`
- `POST /api/v1/prs/{id}/merge`
- `GET /api/v1/runtime/terminal` (no-op; phase 4)
- `GET /api/v1/observability`
- `GET /api/v1/orchestrators`
- `GET /api/v1/version`
- `GET /api/v1/verify`
- `POST /api/v1/update`
- `POST /api/v1/setup-labels`
- `POST /api/v1/webhooks/*` (wildcard)
- `GET /api/v1/filesystem/browse`
- `GET /api/v1/events` (SSE placeholder; phase 2)
- `GET /mux` (placeholder; phase 4) — mounted outside `/api/v1`

Each will get the same REST audit treatment as #20 / #21 once the TS source is read. Expected fixes:
- Errors → locked envelope `{error, code, message, requestId}`.
- Inconsistent success flags dropped.
- Legacy paths documented via `planned.legacy`, not registered.

## Exit criteria

All remaining routes return `501 + planned envelope`. `GET /mux` and `GET /api/v1/events` skip the REST-timeout group (long-lived / streaming).

---
Closes part of #18.

neversettle17-101:
Merged into #21. The tentative route list here has been reconciled against the decided [route analysis](https://neversettle17-101.github.io/ao-backend-arch-design/ao-routes-analysis.html): most of these routes resolve to the **SCM lane** (issues, backlog, verify, prs/merge, webhooks, setup-labels) or are **dropped** (version, update, browse-directory) and are no longer part of the daemon REST shell. The remaining build-now shells (orchestrators, observability, filesystem/browse, events/mux placeholders) now live in #21's combined route surface.

Issue #21, opened 2026-05-28
feat(api): 1b — PR action route shell (merge + resolve-comments)


neversettle17-101:
Part of #10 — Go HTTP Daemon lane. Sub-issue of umbrella #18.

> ⚠️ **Scope:** Interface-first. Register the two SCM/PR action routes as `200` with a structured `planned` body. **No business logic. No real service impls.** Route surface locked to the [SCM section of the route analysis](https://neversettle17-101.github.io/ao-backend-arch-design/ao-routes-analysis.html#scm).

> 🔀 Supersedes #22.

## Depends on
- #14 (Skeleton).
- #20 (#18a Projects) — lands shared scaffolding: `httpd/api.go`, `notimpl.go`, `errors.go`.

## Route surface (2 routes)

Both carry the **different lane** verdict — they belong to the SCM poller / webhook layer (feeding `ApplySCMObservation`), not the session/project REST API. Registered here as shells only.

| Method | Path | Verdict | Notes |
|---|---|---|---|
| POST | `/api/v1/prs/{id}/merge` | different lane | Merge a PR. |
| POST | `/api/v1/prs/{id}/resolve-comments` | different lane | Resolve review comments on a PR. |

## Contract details

**`POST /api/v1/prs/{id}/merge`**
- Request: _(none)_
- `200` → `{ ok: true, prNumber, method: "squash" }`
- `404` → unknown PR
- `409` → not mergeable
- `422` → merge preconditions unmet

**`POST /api/v1/prs/{id}/resolve-comments`**
- Request: `{ commentIds?: string[] }` — omit to resolve all
- `200` → `{ ok: true, resolved: number }`
- `404` → unknown PR
- `422` → nothing to resolve

## Files

```
backend/internal/
  httpd/controllers/
    prs.go            NEW
    prs_test.go       NEW
```

## Exit criteria

```bash
curl -i -XPOST localhost:3001/api/v1/prs/1/merge              # 200
curl -i -XPOST localhost:3001/api/v1/prs/1/resolve-comments   # 200
```

---
Closes part of #18.

Issue #20, opened 2026-05-28
feat(api): 1b·a — Projects route shell (7 routes, REST-corrected)


neversettle17-101:
Part of #10 — Go HTTP Daemon lane. Sub-issue of umbrella #18.

> ⚠️ **Scope:** Interface-first. Register 7 canonical project routes as `501 Not Implemented` with a structured `planned` body. **No business logic. No real `ProjectService` impl.**

## Depends on
- #14 (Skeleton PR for #10·1a) must be merged first.
- Sibling: #18b (Sessions), #18c (Issues+rest), #19 (OpenAPI follow-up).

## Goal

All 7 canonical project routes mounted under `/api/v1` with the chi REST timeout group from the skeleton. Each handler returns `501` + `{error, code, message, requestId, planned}`. Legacy TS paths are NOT registered — see REST audit below.

## REST audit (applies across #18a/b/c)

Reading the old TS surface at `packages/web/src/app/api/**` revealed REST violations. The Go shell **corrects** them and documents the legacy mapping in `planned.legacy` + handler comments, instead of registering aliases.

| # | TS violation | Go fix |
|---|---|---|
| R3 | `PUT /projects/:id` aliased to `PATCH` | Keep PATCH only. PUT not registered — chi returns 405. |
| R4 | `POST /projects/:id` overloaded for repair | Canonical `POST /projects/{id}/repair`. Legacy path not registered. |
| R5 | Degraded `GET /:id` returns 200 with `error` field | Discriminator `{project, status:"ok"\|"degraded"}`. |
| R6 | Inconsistent `ok`/`success` flags | Drop on 2xx. Return affected resource. |
| R9 | Bare `{error: msg}` | Locked envelope `{error, code, message, requestId, details?}`. |

## Route surface (7 canonical)

| Method | Path | Notes |
|---|---|---|
| GET | `/api/v1/projects` | `200 {projects: ProjectSummary[]}` |
| POST | `/api/v1/projects` | Add. `201 {project}`. `400 INVALID_JSON\|PATH_REQUIRED\|NOT_A_GIT_REPO`, `409 PATH_ALREADY_REGISTERED\|ID_ALREADY_REGISTERED` with `details.existingProjectId,suggestedProjectId`. |
| GET | `/api/v1/projects/{id}` | `200 {project, status:"ok"\|"degraded"}`, `404 PROJECT_NOT_FOUND`. |
[...]

Issue #19, opened 2026-05-28
feat(api): 1c — OpenAPI 3.1 spec + request/response validation


neversettle17-101:
Part of #10 — Go HTTP Daemon lane. Follow-up to #18.

> ⚠️ **Scope:** Spec + validation only. Builds on the route shell from #18; does not change route surface or add business logic.

## Goal

Formalise the 25-route contract registered in #18 as a single OpenAPI 3.1 document, and wire a validation middleware so every request and response is checked against the spec at runtime. This locks the contract before phase 1c handlers land.

**Depends on:** #18 (Route shell) must be merged first — the contract-preview bodies in each 501 stub are the source material that gets translated into the spec.

---

## Tasks

- [ ] Author \`backend/api/openapi.yaml\` (OpenAPI 3.1) covering all 25 routes from #18 — paths, methods, params, request bodies, response schemas, status codes
- [ ] Pick & wire a validator: \`github.com/getkin/kin-openapi/openapi3filter\` (recommended — pure runtime validation, no codegen lock-in) vs \`oapi-codegen\` (codegen path)
- [ ] Add validation middleware on the \`/api/v1\` subrouter: rejects malformed requests with \`400\` + structured error, and (in dev/test mode) flags non-conformant responses
- [ ] Serve the spec at \`GET /api/v1/openapi.yaml\` and \`GET /api/v1/openapi.json\` for client consumers
- [ ] Conformance test: table-driven test that loads the spec, hits every route with a minimal valid request, asserts the response validates against the documented schema (even though body is still 501-with-contract-preview)
- [ ] Document the spec authoring workflow in \`backend/api/README.md\` (how to add a route, how to regen if codegen is chosen, how validation errors surface)

## Inputs to the spec

- Contract previews embedded in each #18 controller stub (the \`planned.request\` / \`planned.response\` bodies)
- Prior TS orchestrator contracts (links pending from owner)
- New LCM / SCM port contracts in \`backend/internal/ports/\`

## Design notes / open questions

- **Validator choice:** \`openapi3filter\` keeps the spec as source-of-truth without generated handler glue. \`oapi-codegen\` generates strongly-typed handler interfaces but couples the handler signature to spec regen. Pick \`openapi3filter\` unless the team explicitly wants codegen.
- **Response validation in prod:** validating responses on every request is expensive — gate behind a config flag (\`HTTPD_VALIDATE_RESPONSES\`, default off in prod, on in tests).
[...]

PR #24, opened 2026-05-28
feat(api): projects route shell (7 routes, REST-corrected) — #20

neversettle17-101:
Implements #20 (#18a) — Projects route shell + request/response transport on top of the Phase 1a skeleton (#14). Rebased on `main`.

**Design reference:** [ao-routes-analysis](https://neversettle17-101.github.io/ao-backend-arch-design/ao-routes-analysis.html) — the legacy route audit + per-route verdicts this PR implements.

## Scope

Maps the `/api/v1` projects layer: routes, request decoding, response encoding, the OpenAPI contract, and the package boundaries. **No `project.Manager` implementation** — that lands on top of this. With the (nil) Manager the route shell returns `501`; with a Manager injected the same handlers run end-to-end and a correct request returns `2xx` (verified below).

This PR also lands the **code-first OpenAPI + TypeScript codegen pipeline** (Go is the source of truth; `openapi.yaml` and the frontend types are generated, with a CI drift guard) — see the OpenAPI section below.

## Route surface (5 canonical)

| Method | Path | Success | Errors |
|---|---|---|---|
| GET | `/api/v1/projects` | `200 {projects: Summary[]}` | `500` |
| POST | `/api/v1/projects` | `201 {project}` | `400 INVALID_JSON\|PATH_REQUIRED\|NOT_A_GIT_REPO`, `409 PATH_ALREADY_REGISTERED\|ID_ALREADY_REGISTERED` (`details.suggestion`), `500` |
| GET | `/api/v1/projects/{id}` | `200 {status:"ok"\|"degraded", project}` | `404 PROJECT_NOT_FOUND`, `500` |
| PATCH | `/api/v1/projects/{id}` | `200 {project}` | `400 INVALID_JSON\|IDENTITY_FROZEN\|INVALID_LOCAL_CONFIG`, `404`, `409 PROJECT_DEGRADED\|PROJECT_MISSING_PATH`, `500` |
| DELETE | `/api/v1/projects/{id}` | `200 {projectId, removedStorageDir}` | `400 INVALID_PROJECT_ID`, `404`, `500` |

Per the route-analysis verdicts, `POST /projects/reload` is **dropped** (a Next.js in-process cache hack) and `POST /projects/{id}/repair` is **deferred**. Legacy `PUT /projects/{id}` and `POST /projects/{id}` are unregistered → `405`.

## REST corrections vs legacy TS surface

| # | TS violation | Go fix |
|---|---|---|
| R3 | `PUT /projects/{id}` aliased to PATCH | PATCH only; PUT → 405 |
| R5 | degraded GET returns 200 with an `error` field | discriminator `{status, project}` |
| R6 | inconsistent `ok`/`success` flags | dropped on 2xx; return the affected resource |
| R9 | bare `{error: msg}` | locked envelope `{error, code, message, requestId, details?}` |

## Request ↔ response layer (where the shapes live)

Split by *product command* vs *HTTP wire shape* (`backend-code-structure.md`):

| Piece | Lives in | Why |
|---|---|---|
| Request commands — `AddInput`, `UpdateConfigInput` | `internal/project/dto.go` | Product commands shared by HTTP + CLI; handlers decode the body straight into them |
| Response envelopes — `{projects}`, `{project}`, `{status, project}` | `internal/httpd/controllers/dto.go` | HTTP-only wire shapes; `getProjectResponse` maps `project.GetResult` onto the OpenAPI `oneOf` |
| Transport primitives — JSON writer, error envelope, body decode | `internal/httpd/httpx` | shared leaf package so both `httpd` and `controllers` use it without an import cycle |

Each handler runs `nil-guard → decode → call Manager → encode`. Only `domain.ProjectID` stays in `domain/`; the controller depends solely on `project.Manager`.

## OpenAPI — code-first, Go is the source of truth

`openapi.yaml` is **generated from Go** (do not hand-edit) and stays OpenAPI `3.1.0`:

- **`apispec.Build()`** (`internal/httpd/apispec/build.go`) reflects the contract types + an operation registry into the document via `swaggest/openapi-go`; **`cmd/genspec`** writes it on `go generate`; `apispec` embeds/serves it at `GET /api/v1/openapi.yaml` and slices it into the 501 bodies as before.
- **Annotations live in two places only:** schema facets (`description`/`enum`/`default`) as struct tags on the existing `project.*` / `httpx.Error` types; operation metadata (paths, status codes, summaries, the `ok|degraded` `oneOf`) in `Build()`. `required` is derived from the `omitempty` convention by an `InterceptProp` hook — structs stay clean. Clean schema/TS names (`Project`, `Summary`, `APIError`, …) via `InterceptDefName`.
- **Frontend types fall out of the same spec:** `openapi-typescript` generates `frontend/src/api/schema.d.ts`; `openapi-fetch` gives a typed client (`src/api/client.ts`). No hand-maintained DTOs on either side.
- **Drift can't sneak in:** Go tests assert `embedded openapi.yaml == fresh Build()` + determinism + route↔spec parity; CI `gen-verify` regenerates both artifacts and `git diff --exit-code`s. (`#19` still owns runtime request/response **validation** against this doc.)

Verified: `go generate` + `npm run gen:api` are deterministic and no-diff on a clean tree; spec parses as `3.1.0`; required arrays match the prior hand-written spec exactly.

## Testing

Route-shell coverage is complete — happy path, body validation, error mapping, and the 501/405/404 structural behaviour:

| Area | Test | What it locks |
|---|---|---|
| Happy path (2xx) | `List_OK`, `Add_Created`, `Get_Discriminator` (ok + degraded), `UpdateConfig_OK`, `Remove_OK` | each endpoint's success status + body shape; request decoded into the right command (Manager wired) |
| Body validation (400) | `BodyValidation` — POST & PATCH × {malformed, empty} | `400 INVALID_JSON` + full envelope; **Manager is not called** when decode fails |
| Error mapping (500) | `ManagerError` — all 5 endpoints | each route's documented 500 code, matching the spec's 500 examples |
| Route shell (no Manager) | `Canonical501` (all 5 → 501 + `spec` slice, right `operationId`) | the 501 stub envelope and that each route maps to its OpenAPI operation |
| Legacy / missing | `LegacyUnregistered` (PUT/POST → 405, repair → 404), `MissingRoute` (404 JSON) | dropped verbs and unknown paths return the right status + envelope |
| Contract published | `OpenAPIYAMLServed` + `apispec` unit tests | spec served at `/api/v1/openapi.yaml`; embed loads, operation lookup, param inheritance |

Every non-2xx assertion runs through one `assertEnvelope` helper that pins R9: `error`, `code`, `message`, and a correlation `requestId` are all present.

## Verification

```
go build ./...   ✓     go vet ./...   ✓     go test ./...   ✓     gofmt -l   clean
```

**Correct request → 2xx** (Manager wired with a fake; the impl PR drops in the real one):

```
GET    /api/v1/projects          -> 200  {"projects":[{"id":"demo","name":"Demo","sessionPrefix":"demo"}]}
POST   /api/v1/projects          -> 201  {"project":{"id":"demo","name":"Demo","path":"/tmp/repo",...}}
GET    /api/v1/projects/demo     -> 200  {"status":"ok","project":{"id":"demo",...}}
PATCH  /api/v1/projects/demo     -> 200  {"project":{"id":"demo",...,"agent":{"default":"claude"}}}
DELETE /api/v1/projects/demo     -> 200  {"projectId":"demo","removedStorageDir":true}
```

**Route shell (nil Manager) + envelopes:**

```
curl -i localhost:3001/api/v1/projects                 # 501 + spec slice
curl -i -XPUT  localhost:3001/api/v1/projects/p1        # 405 (R3)
curl -i -XPOST localhost:3001/api/v1/projects/reload    # 405 (dropped)
curl -i -XPOST localhost:3001/api/v1/projects/p1/repair # 404 (deferred)
curl -i        localhost:3001/api/v1/openapi.yaml        # 200 full spec
curl -i        localhost:3001/api/v1/missing             # 404 JSON envelope
```

## Out of scope

- Real `project.Manager` implementation + persistence → handler-impl PR (builds on this).
- Typed `4xx` error taxonomy (currently any Manager error → documented `500`) → handler-impl PR.
- OpenAPI request/response **validation** middleware → #19.
- Sessions / issues / remaining routes → #21 / #22.

---

Closes part of #18 (umbrella). Refs #20.


backend/internal/domain/config_types.go
neversettle17-101, line 1: Why are these changes required at the moment?

backend/internal/domain/project.go
neversettle17-101, line 1: There are project objects defined already in project/ . how do we differentiate it

backend/internal/domain/config_types.go
neversettle17-101: Good catch — they weren't all needed. Auditing references showed `AgentConfig`, `AgentPermission`, `RoleAgentConfig`, and `LocalProjectConfig` had **zero references** anywhere (dead code), and the `Extra map[string]any` passthrough fields had no marshaller, so they silently dropped unknown keys (premature).

Fixed in 9a0547e: deleted `domain/config_types.go`, kept only the 4 shapes the projects API actually exposes (`TrackerConfig`, `SCMConfig`, `SCMWebhookConfig`, `ReactionConfig`) and moved them into `project/types.go` next to the entities that use them. The full Zod port / passthrough round-trip will land with the handler-impl + SQLite work, not this interface-only PR.

backend/internal/domain/project.go
neversettle17-101: Agreed, the split was confusing. Root cause: project types lived in two places — entities in `domain/project.go`, the Manager + DTOs in `project/`.

Fixed in 9a0547e by consolidating **all** project types into the `project` package:
- `domain/project.go` → `project/types.go`: `Project`, `Summary`, `Degraded` (dropped the `Project` prefix since the package name carries it — `project.Summary`, `project.Degraded`).
- Only `domain.ProjectID` stays in `domain`, because it's shared vocabulary referenced by sessions/lifecycle/workspace/ports.

So now there's one home for everything project-shaped (entities in `types.go`, interface in `project.go`, DTOs in `dto.go`), and `domain/` no longer carries project-specific types. This is the feature-package layout we're standardizing on.

illegalcall (commented):
Structure feedback, using the package-ownership rules documented in #41 as the reference point.

Direct take: the direction is good, but I would tighten a few boundaries before using this PR as the template for the next resources.

1. `internal/project` is the right home for project-owned concepts. Keeping `Project`, `Summary`, `Degraded`, project behavior config shapes, and project command/result types together is better than putting them in `domain`, `ports`, or `httpd`. This is the right feature-package direction.

2. `domain` should stay smaller than this PR implies in comments. `domain.ProjectID` is appropriate because sessions/lifecycle/workspace need the same identity vocabulary. Full project read models and project config shapes should stay in `internal/project`, as this PR does.

3. The `project.Manager` boundary is okay only if it is meant to be an application-service contract used beyond HTTP, for example HTTP + CLI. If it is only a controller seam, I would define the small interface in `httpd/controllers` and let `project` expose a concrete service later. Rule: reusable product workflow contracts live in the feature package; protocol-only seams live with the protocol consumer.

4. `project.GetResult` needs a clearer DTO/wire boundary. The comment says `Status` mirrors the wire discriminator, but the struct has no JSON tags and has `Project` plus `Degraded`, while OpenAPI defines `{ status, project }` where `project` is oneOf Project/DegradedProject. Either make `GetResult` an internal app result and map it in `httpd`, or make it explicitly match the OpenAPI wire shape. Do not leave it halfway between both.

5. The comments in `controllers/projects.go` still say managers come from `ports/inbound.go`. That is stale now that the controller depends on `project.Manager`.

6. Integration note: this PR targets `feat/issue-10`; against current `main`, there are conflicts in `backend/go.mod`, `backend/go.sum`, and `backend/internal/httpd/router.go`. Since the package structure is the part we want to reuse, I would refresh it against current `main` before treating it as settled.

Overall: +1 on `internal/project` as the pilot feature package and +1 on keeping OpenAPI/HTTP concerns under `httpd`. The main thing I would change is the phrasing/pattern around “one Manager per resource” so it does not become another central-ports pattern, just distributed by package name.

backend/internal/project/dto.go
```diff
@@ -0,0 +1,55 @@
+package project
+
+import "github.com/aoagents/agent-orchestrator/backend/internal/domain"
+
+// Request/response shapes for Manager. They carry the data across the service
+// boundary; the entities they reference (Project, Summary, Degraded) live in
+// types.go in this same package. Named without a "Project" prefix because the
+// package name already supplies it (project.AddInput, project.GetResult).
+
+// GetResult is the discriminated union returned by Manager.Get. Exactly one of
```
illegalcall: Clarification: is `GetResult` intended to be an internal application result, or the HTTP wire response? The comment says it mirrors the wire discriminator, but this struct does not match the OpenAPI shape exactly. I would prefer making this an internal app result and mapping it in `httpd`, unless we intentionally want `project` to own the wire shape.

backend/internal/project/dto.go
```diff
@@ -0,0 +1,55 @@
...
+// Project / Degraded is non-nil; Status mirrors the discriminator on the wire
+// so consumers branch on it without nil-checking both fields.
+type GetResult struct {
```
illegalcall, line 12: If this remains an HTTP-facing DTO, it should probably have explicit JSON tags and match the documented response shape. OpenAPI has `{ status, project }`, where `project` is oneOf Project/DegradedProject; this struct has separate `Project` and `Degraded` fields. That makes the boundary a little unclear.

backend/internal/project/project.go
```diff
@@ -0,0 +1,44 @@
...
+// later handler-impl PR. This PR defines only the contract.
+package project
+
+import (
+	"context"
+
+	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
+)
+
+// Manager is the inbound contract for the /api/v1/projects surface. One
```
illegalcall: Clarification: I think `project.Manager` belongs here if it is the reusable application-service contract, for example something both HTTP and CLI will call. If this interface only exists for the HTTP controller seam, I would move the interface closer to the consumer in `httpd/controllers` and let `project` expose the concrete service later.

backend/internal/httpd/controllers/projects.go
```diff
@@ -0,0 +1,78 @@
+// Package controllers holds the HTTP-facing controllers for the /api/v1
+// surface. Each controller groups one resource's routes, exposes a Register
+// method that wires them on a chi.Router, and depends on exactly one
+// *Manager interface from ports/inbound.go — never on a store, the LCM, an
```
illegalcall: This looks stale after moving the project contract into `internal/project`: this now depends on `project.Manager`, not `ports/inbound.go`.

backend/internal/httpd/apispec/openapi.yaml
```diff
@@ -0,0 +1,447 @@
...
+    DegradedProject:
+      type: object
+      required: [id, name, path, resolveError]
+      properties:
+        id: { type: string }
+        name: { type: string }
+        path: { type: string }
+        resolveError: { type: string }
+
+    ProjectGetResponse:
```
illegalcall, line 290: This is the shape I would treat as HTTP-owned. If `project.GetResult` stays different from this, the controller should do an explicit mapping from the project application result into this response envelope.

illegalcall (commented):
Adding a few line-specific structure clarifications so the package boundary is easier to settle before later resources copy this pattern.

backend/internal/project/dto.go
neversettle17-101: Resolved in 518097e by committing to one side of the boundary: `GetResult` is now documented as an **internal application result** (no JSON tags), and the httpd controller maps it onto the `ProjectGetResponse` wire envelope. The project package keeps the ergonomic two-pointer shape; httpd owns the OpenAPI `oneOf` representation.

backend/internal/project/dto.go
neversettle17-101: Done — went with the internal-result option rather than wire-matching. `GetResult` stays tag-free with `Project`/`Degraded` pointers; the comment now states explicitly that httpd maps it to `{ status, project }` (project = oneOf Project/DegradedProject). Mapping lands with the handler-impl PR; no halfway DTO. (518097e)

backend/internal/project/project.go
neversettle17-101: Confirmed intent: `Manager` is an application-service contract meant for HTTP **and** CLI (add/remove/reload/repair are reusable product workflows), mirroring `ports.SessionManager` — not a controller-only seam. Tightened the package + interface docs in 518097e to say so, so it stays in `internal/project` per the rule.

backend/internal/httpd/controllers/projects.go
neversettle17-101: Good catch — stale. Fixed in 518097e: the comment now says the controller depends on the feature package's application-service contract (`project.Manager` in `internal/project`), not `ports/inbound.go`.

backend/internal/httpd/apispec/openapi.yaml
neversettle17-101: Agreed — this shape is HTTP-owned. `project.GetResult` is now an internal result, and the controller will do the explicit mapping into this `ProjectGetResponse` envelope when handlers land. Documented that contract on both sides in 518097e.

backend/internal/project/dto.go
neversettle17-101: Resolved in 518097e: committed to the internal-result side. `GetResult` is now documented as an **internal application result** (no JSON tags); the httpd controller maps it onto the `ProjectGetResponse` wire envelope. `project` keeps the ergonomic two-pointer shape; httpd owns the OpenAPI `oneOf`.

backend/internal/project/dto.go
neversettle17-101: Done in 518097e — went internal rather than wire-matching. `GetResult` stays tag-free with `Project`/`Degraded` pointers; the comment now states httpd maps it to `{ status, project }` (project = oneOf Project/DegradedProject). No halfway DTO.

backend/internal/project/project.go
neversettle17-101: Confirmed: `Manager` is a reusable application-service contract for HTTP **and** CLI (add/remove/reload/repair are product workflows), mirroring `ports.SessionManager` — not a controller-only seam. Tightened the package + interface docs in 518097e to say so; it stays in `internal/project`.

backend/internal/httpd/controllers/projects.go
neversettle17-101: Stale, fixed in 518097e: the comment now references the feature-package contract (`project.Manager` in `internal/project`), not `ports/inbound.go`.

backend/internal/httpd/apispec/openapi.yaml
neversettle17-101: Agreed — HTTP-owned. `project.GetResult` is now an internal result; the controller does the explicit mapping into this `ProjectGetResponse` envelope when handlers land. Documented on both sides in 518097e.

backend/internal/httpd/apispec/apispec.go
```diff
@@ -0,0 +1,153 @@
...
+		Code:      "NOT_IMPLEMENTED",
+		Message:   method + " " + path + " is registered but not yet implemented",
+		RequestID: middleware.GetReqID(r.Context()),
+		Spec:      op,
+	}
+	w.Header().Set("Content-Type", "application/json; charset=utf-8")
+	w.WriteHeader(http.StatusNotImplemented)
+	// A write error here means the client went away mid-response.
+	_ = json.NewEncoder(w).Encode(body)
+}
```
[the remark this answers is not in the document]
neversettle17-101: Fixed in 6343478. A missing spec slice is a server-side defect (route wired without OpenAPI coverage), not a "not implemented" condition — so `NotImplemented` now logs and returns `500` with code `INTERNAL` (not `NOT_IMPLEMENTED`) via the locked error envelope, and only emits the 501 when the slice is present. Also dropped the misleading `omitempty` on `spec`. Covered by new tests for both the 500 fallback and the known-route 501 still carrying its slice.

AgentWrapper:
Heads-up on the conflicts here: they're caused by **#47** ("implement project routes with mock manager/store"), which squash-merged to `main` and independently shipped the same projects-routes feature this PR builds — hence the `add/add` conflicts on the `project` package, `controllers/projects.go`, `api.go`, and `apispec/*`. The route shell itself is now redundant with what #47 put on `main`.

Two parts of this PR, though, are genuinely net-new and **not** on `main`: the **code-first OpenAPI generation** (`cmd/genspec` + `apispec` build/parity, replacing the hand-maintained `openapi.yaml`) and the **typed frontend client** (`frontend/src/api/`). Rather than resolve a ~13-file `add/add` conflict (which would risk reverting #47's merged work), I've salvaged just those two pieces, rebuilt on top of #47's code, in a fresh PR:

👉 **#59 — https://github.com/aoagents/agent-orchestrator/pull/59**

It's green (backend build/vet/test incl. drift + parity tests, frontend gen + typecheck). One tradeoff called out there: regenerating the spec drops the hand-authored examples/`x-rest-audit` notes from #47's `openapi.yaml` — happy to re-add those as `specgen` operation metadata if you'd prefer.

Suggest **closing this PR as superseded by #47 (+ #59 for the codegen/frontend delta)** — but it's your call, and I didn't want to touch it without flagging you first. 🙏

neversettle17-101:
> Heads-up on the conflicts here: they're caused by **#47** ("implement project routes with mock manager/store"), which squash-merged to `main` and independently shipped the same projects-routes feature this PR builds — hence the `add/add` conflicts on the `project` package, `controllers/projects.go`, `api.go`, and `apispec/*`. The route shell itself is now redundant with what #47 put on `main`.
> 
> Two parts of this PR, though, are genuinely net-new and **not** on `main`: the **code-first OpenAPI generation** (`cmd/genspec` + `apispec` build/parity, replacing the hand-maintained `openapi.yaml`) and the **typed frontend client** (`frontend/src/api/`). Rather than resolve a ~13-file `add/add` conflict (which would risk reverting #47's merged work), I've salvaged just those two pieces, rebuilt on top of #47's code, in a fresh PR:
> 
> 👉 **#59 — #59**
> 
> It's green (backend build/vet/test incl. drift + parity tests, frontend gen + typecheck). One tradeoff called out there: regenerating the spec drops the hand-authored examples/`x-rest-audit` notes from #47's `openapi.yaml` — happy to re-add those as `specgen` operation metadata if you'd prefer.
> 
> Suggest **closing this PR as superseded by #47 (+ #59 for the codegen/frontend delta)** — but it's your call, and I didn't want to touch it without flagging you first. 🙏

The #47 PR was built up on top the current PR's previous commits. 
1. There are certain routes implemented in #47 which are not needed.
2. There are redundant files backend/internal/project/errors.go which should ideally reuse the http/ folder one. 
3. The #47 PR implements [manager](backend/internal/project/manager.go) but doesn't reuse the existing interface of backend/internal/project/project.go. 
4. There is a new backend/internal/project/memory_store.go created inside the projects/ folder which is just extra code and not required and should be added inside the projectManager only. The comments for the memory_store.go say `// MemoryStore is the mocked DB layer for the project API implementation.` 
5. I do not see the testing details of the routes implemented with service logic in #47. I do not see the connection with the storage layer in the PR merged

I was planning to work on all these items and add an end to end description testing suite as well. Open to close this PR if we are okay with the above mentioned points.

neversettle17-101:
> 1. There are certain routes implemented in [feat(api): implement project routes with mock manager/store #47](https://github.com/aoagents/agent-orchestrator/pull/47) which are not needed.
> 2. There are redundant files backend/internal/project/errors.go which should ideally reuse the http/ folder one.
> 3. The [feat(api): implement project routes with mock manager/store #47](https://github.com/aoagents/agent-orchestrator/pull/47) PR implements [manager](backend/internal/project/manager.go) but doesn't reuse the existing interface of backend/internal/project/project.go.
> 4. There is a new backend/internal/project/memory_store.go created inside the projects/ folder which is just extra code and not required and should be added inside the projectManager only. The comments for the memory_store.go say `// MemoryStore is the mocked DB layer for the project API implementation.`
> 5. I do not see the testing details of the routes implemented with service logic in [feat(api): implement project routes with mock manager/store #47](https://github.com/aoagents/agent-orchestrator/pull/47). I do not see the connection with the storage layer in the PR merged
> 
> I was planning to work on all these items and add an end to end description testing suite as well. Open to close this PR if we are okay with the above mentioned points.

@AgentWrapper  I will address these comments in a new PR. Closing the existing one.

Issue #10, opened 2026-05-27
Tracking: Go HTTP Daemon — REST


neversettle17-101:
> ⚠️ **Disclaimer:** This issue is primarily focused on **defining interfaces and service boundaries**. Implementation details, data models, and internal wiring are intentionally pushed to later phases — the goal right now is to agree on the shape of the system before writing any code.

---

Tracking issue for the **Go HTTP Daemon** lane of the backend rewrite.

## Goal

Replace the current Next.js API routes with a single Go binary on `127.0.0.1:3001` that serves:
- **`/api/v1/*`** — REST endpoints (25 routes, path-parity with current TS)

The daemon is spawned as a **sidecar child process** by the Electron main process, supervised over loopback only. No auth, no CORS, no public URL.

## Approach: phase-by-phase, contract-first

`feat/go-http-daemon` is the **lane integration branch**: every sub-PR branches off it and merges **into** it; the whole lane lands on `main` as one unit once complete.

Design doc & HTML overview: https://htmlpreview.github.io/?https://gist.githubusercontent.com/neversettle17-101/00c659adf978b80b1e2eb600d451932a/raw/ao-backend-rewrite-phase1.html

---

## Startup sequence

```
1. Electron main spawns ./ao-daemon
2. Daemon binds :3001, starts listening
3. Main polls /healthz until ready
4. Renderer connects and loads UI
```

> **TODO — open question:** How are agent sessions & projects loaded on daemon startup?
> Warm cache from disk, lazy-load on first request, or explicit restore step?

---

### Phase 1 — HTTP server skeleton + Electron integration

- [x] **1a — Skeleton** — Go module init, `config` (port, env), `127.0.0.1` bind, 5-middleware stack (recoverer → request ID → logger → real IP → timeout), `/healthz` + `/readyz`, graceful shutdown (10s timeout, SIGTERM/SIGINT)
- [ ] **1b — Route shell** — All 25 routes registered under `/api/v1` as `501 Not Implemented` stubs; subrouter layout (`/events`, `/mux`, `/api/v1`, `/`) with correct middleware applied per surface
- [ ] **1c — Electron integration** — `embed.FS` static serving, Electron main process spawns + supervises daemon, health-check polling before renderer connects, dev proxy (Next.js `:3000` → Go `:3001`), packaged build with `extraResources`

- [ ] **Sessions** — list (with `?project`, `?active`, `?orchestratorOnly`, `?fresh`), get, spawn, kill, message, send, restore
[...]

Issue #9, opened 2026-05-27
Tracking: LCM+SM runner — reaper, daemon wiring, Runtime/Workspace adapters


harshitsinghbhandari:
Follow-on to #3 (the LCM + Session Manager lane core, merged via #2). That core is implemented and tested **behind fakes**; this issue tracks the **runner + the adapters we own** to get from "tested behind fakes" to a daemon that actually spawns and supervises sessions.

## Owner
@harshitsinghbhandari (+ @yyovil on the `Agent` adapter — a dependency, built separately).

### Runner / wiring
- [ ] **Reaper** — a periodic timer that (a) samples runtime/process liveness per active session → `LCM.ApplyRuntimeObservation`, and (b) calls `LCM.TickEscalations(now)` so duration-based escalations actually fire. (The design assigns the reaper to this lane; the LCM never polls itself.)
- [ ] **Daemon wiring** — replace the `main.go` stub with the real assembly: construct the LCM + Session Manager with their adapters, start the reaper loop, and handle graceful start/stop/shutdown.
- [ ] **End-to-end harness** — wire LCM + SM + in-memory fakes and prove a full **spawn → observe → react → kill** flow (integration test), independent of the external lanes.

### Carried-forward items (from the core PRs)
- [ ] **`react()` out-of-lock dispatch ordering** (from #6) — reactions fire after the per-session lock releases; give `react()` a per-session ordering (a small react queue) or re-check the triggering state before dispatch, so concurrent observers can't dispatch on a stale snapshot.
- [ ] **`ExpectedRevision` wiring** (from #5) — use the `LifecyclePatch.ExpectedRevision` optimistic-CAS for multi-writer / CDC safety (currently the only guard is the in-process per-session mutex).

### Real adapters we own
- [ ] **`Runtime` — tmux adapter** — `Create`/`Destroy`/`SendMessage`/`GetOutput`/`IsAlive` (tmux `new-session`/`has-session`/`send-keys`/`capture-pane` mechanics).
- [ ] **`Workspace` — git-worktree adapter** — `Create`/`Destroy`/`List`/`Restore`, honoring the worktree-remove safety (refuse to `rm -rf` a path still registered after `git worktree prune`).

## Dependencies (other lanes — coordinate; not blocking the above)
- **`Agent` adapter** (claude-code / codex) → @yyovil. The reaper's liveness sampling and the SM's spawn use the `Agent` port; our work proceeds against the port + a fake until it lands.
[...]

PR #29, opened 2026-05-29
feat(observe): reaper for liveness probe + TickEscalations heartbeat (#9)

harshitsinghbhandari:
## Summary

Builds the reaper for the OBSERVE layer (issue #9 / aa-20). Scope: reaper goroutine ONLY — no activity ingest, no FS watcher, no daemon wiring, no schema/store changes.

The reaper sits OUTSIDE the LCM's per-session serial loop. Each tick it:

1. Fires `lcm.TickEscalations(now)` — the duration-based escalation heartbeat the synchronous LCM cannot wake itself to drive.
2. Asks `lcm.RunningSessions(ctx)` for sessions still worth probing (non-terminal) and probes each one via `runtime.IsAlive(handle)`, dispatching through a `RuntimeRegistry` keyed by `RuntimeHandle.RuntimeName` (so tmux + zellij share one reaper).
3. Reports the probe result back via `ApplyRuntimeObservation` — alive / dead / failed all flow through. The reaper does NOT decide what counts as a no-op or what counts as "truly dead"; the LCM owns DECIDE (anti-flap Detecting quarantine, terminal-session rules).

A probe error is reported as a `RuntimeProbeFailed` fact, never collapsed to alive or dead — the LCM's "failed probe ≠ dead" invariant holds.

## Open question — resolution

The spec listed three options for how the reaper enumerates sessions:
- (a) Add `LifecycleManager.RunningSessions` — read snapshot via LCM
- (b) Inject `LifecycleStore` directly
- (c) Subscribe/changes channel

**Chose (a).** Reasoning:
- Keeps the reaper's dependency surface to a single port (`LifecycleManager`).
- "Worth probing" is a lifecycle predicate the LCM should own and can evolve (e.g. exclude in-cleanup, include not_started) without touching the reaper.
- Avoids forcing a `LifecycleStore` contract change (Tom-store untouched per spec). The Manager implements `RunningSessions` via an injectable `sessionLister` (`Manager.WithSessionLister`); daemon wiring (#10) will hook this to the store at startup.

### Predicate detail (worth highlighting for review)

Initial draft filtered to `runtime.State == RuntimeAlive`. Self-review caught that this silently parks every Detecting session — a session enters Detecting with runtime axis = `RuntimeProbeFailed` (failed-probe path, `decide_bridge.go:72`) or `RuntimeMissing` (`manager_test.go:539`). Both differ from `RuntimeAlive`. That broke:
- The probe-driven recovery path proved by `manager_test.go:59` ("healthy probe recovers a liveness-owned detecting -> working").
- The terminal escalation path proved by `manager_test.go:79` ("dead+dead with no recent activity concludes killed") — only the first probe would ever reach the LCM.

Final predicate: "not in a final session state" (mirrors the LCM's existing `isTerminal` helper). Locked in by `TestRunningSessions_FilterIncludesProbableExcludesTerminal`.

Also dropped a skip-alive optimization in `probeOne`: the LCM already collapses no-op diffs in `ApplyRuntimeObservation`, and dropping the skip is what makes alive-probe recovery from Detecting reachable.

## Seam concerns / known follow-ups

- **Partial-failure log flood (no rate limit).** `Tick` logs every per-session `ApplyRuntimeObservation` error but doesn't back off. If the LCM is unhealthy across N sessions, every tick emits N error logs. Worth a follow-up (counter + summary log).
- **`TickEscalations` errors are logged-and-continue.** Right shape for steady-state, but no visibility metric. Worth a follow-up.
- **Lifecycle metadata-key constants reach across packages.** `observe/reaper` imports `lifecycle` only for `MetaRuntimeHandleID` / `MetaRuntimeName`. Logically those describe the persistence contract and belong in `ports`. Worth a small follow-up (would also let the reaper depend only on `ports`).
- **`WithSessionLister` / `MapRegistry` are not thread-safe.** Documented as "configure before `Start`". Sufficient for the planned daemon wiring; if a future caller needs dynamic registration, swap to `atomic.Value` / `sync.RWMutex`.
- **Loop test polls with a 500ms window for 2 ticks at 5ms tick.** Race-clean and resilient on a developer laptop / normal CI. If it flakes under load, raise the window.

## What was NOT touched

- Activity ingest, `.ao/activity.jsonl` watcher (separate follow-up).
- Daemon wiring / `main.go` (lane #10 — aditi).
- Any schema fields, any `LifecycleStore` changes.
- Real tmux/zellij — the reaper consumes the existing `IsAlive` adapter contracts via fakes in tests.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test -race ./...` — 250 pass
- [x] Reaper coverage: alive (alive fact reported, no death applied), dead (one `RuntimeProbeDead`), detecting+alive (recovery fact reported), probe error (`RuntimeProbeFailed` fact, not collapsed to alive), multi-runtime dispatch (tmux + zellij probed via correct adapter), unknown runtime skipped, missing handle skipped, goroutine lifecycle (Start/cancel).
- [x] `Manager.RunningSessions` coverage: nil-lister no-op, lister-error propagation, canonical-state matrix include/exclude.

### DIFF
--- a/backend/internal/lifecycle/manager.go
+++ b/backend/internal/lifecycle/manager.go
@@ -77,7 +77,10 @@ func New(store ports.LifecycleStore, notifier ports.Notifier, messenger ports.Ag
 
 // WithSessionLister injects the function the LCM uses to enumerate all
 // persisted sessions for RunningSessions. The daemon wires this against the
-// store at startup. Calling it more than once replaces the previous lister.
+// store at startup; it must be called BEFORE any reaper attached to this
+// Manager starts running, since concurrent calls would race the bare-field
+// read in RunningSessions. Calling it more than once replaces the previous
+// lister.
 func (m *Manager) WithSessionLister(fn func(ctx context.Context) ([]domain.SessionRecord, error)) {
 	m.sessionLister = fn
 }
@@ -426,16 +429,28 @@ func (m *Manager) OnKillRequested(ctx context.Context, id domain.SessionID, r po
 
 // ---- read-snapshot helpers ----
 
-// RunningSessions returns a snapshot of every persisted session whose runtime
-// axis is alive. It exists so the reaper (OBSERVE) can decide whom to probe
-// without taking on a LifecycleStore dependency or knowing the LCM's internal
-// state. Because the call only reads and copies, it does not break the
-// single-writer invariant; concurrent Apply* calls may move sessions in or out
-// of "alive" between snapshots, which is correct — the next tick re-reads.
+// RunningSessions returns a snapshot of every persisted session worth probing
+// in the next reaper tick. "Worth probing" is wider than "runtime axis alive":
+// it includes sessions in the Detecting quarantine, because a fresh probe is
+// the only fact that can recover them (back to working) or escalate them
+// (terminal killed). Filtering to runtime-axis-alive would silently park every
+// Detecting session — a single failed probe would never get a second chance
+// and recovery via runtime probe would be unreachable.
+//
+// The predicate is "not a final session state". Terminal session states (done,
+// terminated) are excluded because Restore is the only path back; observations
+// must not reopen them (#1 invariant). Sessions in earlier states — not_started,
+// working, idle, needs_input, stuck, detecting — are all included. Those that
+// lack runtime handle metadata (e.g. not_started before OnSpawnCompleted) are
+// returned and harmlessly skipped by the reaper's per-session handle guard.
+//
+// The call only reads and copies, so it does not break the single-writer
+// invariant; concurrent Apply* calls may move sessions in or out of the probe
+// set between snapshots, which is correct — the next tick re-reads.
 //
 // When no lister has been wired (e.g. tests construct a bare Manager), the
-// method returns an empty slice so a goroutine attached to such a Manager
-// degrades to a no-op rather than panicking.
+// method returns nil so a goroutine attached to such a Manager degrades to a
+// no-op rather than panicking.
 func (m *Manager) RunningSessions(ctx context.Context) ([]domain.SessionRecord, error) {
 	if m.sessionLister == nil {
 		return nil, nil
@@ -446,7 +461,7 @@ func (m *Manager) RunningSessions(ctx context.Context) ([]domain.SessionRecord, 
 	}
 	out := make([]domain.SessionRecord, 0, len(all))
 	for _, rec := range all {
-		if rec.Lifecycle.Runtime.State == domain.RuntimeAlive {
+		if !isTerminal(rec.Lifecycle.Session.State) {
 			out = append(out, rec)
 		}
 	}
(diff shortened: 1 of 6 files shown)


harshitsinghbhandari (commented):
## Review verdict: **APPROVE-WITH-COMMENTS**

All 8 contract checks pass; build clean, `go test ./...` 250 pass across 13 packages (reaper + lifecycle pass `-race`). Notes file: `~/.me/notes/ao-pr29-review.md`.

### Enumeration choice
Option **(a) `LifecycleManager.RunningSessions` snapshot** — new read-only method on the inbound port (`internal/ports/inbound.go:38`), implemented as a fresh-slice copy in the LCM (`internal/lifecycle/manager.go:454`). Poll-set predicate is "not terminal session state", so Detecting sessions remain probable — this is correct and `TestRunningSessions_FilterIncludesProbableExcludesTerminal` locks the predicate in.

### Contract checks
1. **No store writes** — `grep` over the reaper package returns zero `Upsert`/`LifecycleStore`/`store.` symbol uses. Only `domain`, `lifecycle` (for metadata key constants), and `ports` are imported.
2. **No death-deciding** — `reaper.go:181-197` is a 3-arm switch on `(alive, err)` → translate to facts. No attempts counter, no anti-flap, no terminal inference. The decider stays in `decide_bridge.go` + `decide`.
3. **Enumeration respects (1)** — snapshot via the LCM's public method; iteration is over a value-copy slice, never holds an internal lock.
4. **Multi-runtime dispatch** — per-session lookup via `RuntimeRegistry` (`reaper.go:172`); `multi-runtime dispatch: tmux + zellij in same tick` verifies both adapters get the correct handles and order.
5. **TickEscalations every tick** — fired *before* `RunningSessions` (`reaper.go:140`) and best-effort on error, so the heartbeat always lands. Verified across every table case and `TestReaper_Loop`.
6. **Probe error handling** — `(false, err)` → `RuntimeProbeFailed + ProcessProbeFailed`, never collapsed to alive or dead. Test `probe error: reported as failed fact, NOT collapsed to alive` asserts the exact facts.
7. **Concurrency** — single ticker goroutine, no internal mutexes; `IsAlive` runs without holding any LCM lock; calls back into the LCM via public `Apply*`/`Tick*`/`RunningSessions` only. `-race` clean.
8. **Tests** — table-driven `TestReaper_Tick` covers alive, dead, multi-runtime, probe-error, detecting-recovery; `TestReaper_Loop` exercises ticker + ctx-cancel exit; `SkipsUnknownRuntime` and `SkipsMissingHandle` cover the skip paths; injectable Clock; recording `fakeLCM` asserts exact call sequence.

### Polish (non-blocking)
- `ProcessState` is stamped equal to `RuntimeState` (`reaper.go:191-197`). The reaper only runs `Runtime.IsAlive`; this is consistent with the design giving Activity ingest ownership of agent-process exit. When `ports.Agent.ProbeProcess` is wired by the activity-ingest follow-up, the reaper should probe and report the two axes independently.
- `WithSessionLister` writes the field without locking; docstring says "must be called before any reaper starts" — soft race documented away, fine for v1.
- Per-session probes within a tick are sequential — bounded concurrency would help if the running set grows or any adapter slows down.
- No per-probe timeout independent of ctx; relies on adapter discipline.

Will let harshit merge.

PR #37, opened 2026-05-30
feat(backend): SQLite storage layer + CDC pipeline

Pritom14:
## Summary

Replaces the in-memory `fakeStore` with the two real outbound adapters the Go backend needs to persist and deliver session state:

- **`internal/storage/sqlite`** - persistence satisfying the existing `ports.LifecycleStore` contract.
- **`internal/cdc`** - transactional-outbox publisher, JSONL delivery, durable consumer, and an in-process broadcaster.

Both are wired into `main.go` alongside the Lifecycle Manager and reaper, so the write path is live end-to-end:

```
LCM.Upsert -> sqlite store (1 tx: session + change_log + outbox)
           -> cdc publisher -> session-events.jsonl
           -> cdc consumer (byte cursor, idempotent) -> Broadcaster -> [FE transport, deferred]
```

### Storage (`internal/storage/sqlite`)
- `modernc.org/sqlite` (pure Go, no CGO) so the 3-OS binary cross-compiles cleanly.
- `goose` embedded migrations (`0001_init.sql`); `sqlc`-generated typed queries checked in under `gen/`.
- **Atomic `Upsert`**: the session row, `change_log` entry, and `outbox` row are written in a single `*sql.Tx` (all-or-nothing).
- **`revision` is an optimistic-concurrency (CAS) check**, not a debug counter: insert requires incoming revision `0` and persists `1`; update requires the loaded revision to equal the stored one and bumps it `+1`; zero rows affected returns a revision-mismatch error. This mirrors the original `fakeStore` semantics exactly.
- **Metadata** (`branch`, `workspacePath`, etc.) lives in a separate `session_metadata` table written by `PatchMetadata`, deliberately **off** the canonical CDC path (no revision bump, no event).
- **Durable `reaction_trackers`**: escalation budgets now survive a restart, fixing the in-memory-only behaviour that re-fired human pages after a daemon bounce.

### CDC (`internal/cdc`)
- **Publisher** drains the outbox to a JSONL log; size-based rotation writes a reset marker.
- **Consumer** tails via a byte cursor, detects rotation with `os.SameFile`, resyncs from a full-state snapshot on a gap, and persists its position in `consumer_offsets`.
- **Janitor** reclaims acknowledged outbox rows below the slowest consumer offset.
- **Broadcaster** is the in-process fan-out port the frontend transport will subscribe to (WS/SSE wiring intentionally deferred to the FE task).

### Composition root (`main.go`, `cdc_wiring.go`, `lifecycle_wiring.go`)
- `startCDC` stands up publisher / consumer / janitor + broadcaster.
- `startLifecycle` constructs the LCM, makes escalation budgets durable via `WithReactionStore`, teaches it to enumerate sessions via `WithSessionLister(store.ListAll)`, and starts the reaper.

## Migration 0002: projects + pr_enrichment

Two further tables, both written **outside** the LCM write path (no revision bump, no `change_log`/`outbox` event), so neither rides the CDC pipeline. Store methods follow the existing `reaction_trackers` adapter pattern with storage-local row structs (`ProjectRow`, `PREnrichmentRow`).

### Table: `projects`
The durable registry of repos AO manages, the SQLite twin of the old YAML config (global `config.yaml` plus per-repo `agent-orchestrator.yaml`). Keyed by the same `{basename}_{sha256(path:originUrl)[:10]}` id that `sessions.project_id` references.

- **Function**: persist and enumerate managed repos (path, repo owner/name/platform/origin URL, default branch, display name, session prefix, source, registration time).
- **Soft delete via `archived_at`** (NULL = active). Archiving keeps the row so a session's `project_id` always resolves; there is **no FK** from `sessions.project_id` to `projects.id` (SQLite cannot `ALTER ADD` a FK, and a backfill can land sessions before their project row), so the soft-delete row is what prevents dangling references.
- **Reads**: `ListProjects` returns active rows only (`WHERE archived_at IS NULL`); `GetProject` resolves any row regardless of state and reports `ArchivedAt`.
- **Methods**: `UpsertProject`, `GetProject` (returns `ok`), `ListProjects`, `ArchiveProject`, `DeleteProject` (hard purge).

### Table: `pr_enrichment`
A per-session cache of the rich SCM facts that do **not** belong in the canonical lifecycle (which keeps only `pr_state` / `reason` / `number` / `url`).

- **Function**: store the SCM observer's fetched PR detail: CI summary, review decision, mergeability, pending review comments (opaque JSON blob), CI failure log tail, last-fetched time.
- **1:1 with a session**, `session_id PRIMARY KEY REFERENCES sessions(id) ON DELETE CASCADE`, so enrichment dies with its session (same cascade as `session_metadata`).
- **Methods**: `UpsertPREnrichment`, `GetPREnrichment` (returns `ok`), `DeletePREnrichment`.

Both tables are currently **dormant**: the SCM observer that writes `pr_enrichment` and the project-registration/archive handler that writes `projects` are downstream daemon-lane work and do not exist yet. The schema, queries, and store methods are in place so that wiring is purely additive.

## Temporary stubs and how to fill them in

The write path is live, but three collaborators owned by the daemon/coding-agents lane do not have production implementations yet. They are stubbed with `TODO(daemon-lane)` markers in `lifecycle_wiring.go`. A no-op here is **safe** only because it merely drops an outbound side-effect; none of them fake success.

| Stub | File / call site | How to replace |
|------|------------------|----------------|
| `noopNotifier` | `lifecycle_wiring.go`, passed to `lifecycle.New(store, noopNotifier{}, ...)` | Implement the notifier multiplexer over the real notifier plugins (desktop / Slack / webhook) and pass it in place of `noopNotifier{}`. Until then human-facing notifications are silently dropped. |
| `noopMessenger` | `lifecycle_wiring.go`, passed to `lifecycle.New(store, _, noopMessenger{})` | Back it with the runtime/agent plugins so it injects a prompt into the live agent pane. Until then auto-nudge reactions are no-ops. |
| `reaper.MapRegistry{}` (empty) | `lifecycle_wiring.go`, `reaper.New(lcm, reaper.MapRegistry{}, ...)` | Register the real runtime adapters keyed by name, e.g. `reaper.MapRegistry{"tmux": tmuxRuntime, "process": procRuntime}`. With an empty registry the reaper ticks escalations but probes nothing, which is correct until runtimes exist. |

### Deliberately NOT stubbed (see comment block in `main.go`)
- **Session Manager** - `session.New` needs real `Runtime` / `Agent` / `Workspace` plugins. No-op versions would make `Spawn` *appear* to succeed while creating no live pane (a silent-failure footgun), so it is deferred rather than faked. The LCM already exposes the read surface (`RunningSessions`) the SM would wrap.
- **HTTP API routes** - `httpd.New` takes no SM/LCM today; surfacing the store over HTTP needs a constructor signature change plus handlers that call into the SM, so it is tracked with the SM work.

The distinction: stub when the no-op is semantically honest (no notifier => no notification). Do not stub when the no-op would lie about success (no runtime => ghost session).

## gitignore
Added `*.db`, `*.db-shm`, `*.db-wal`, `session-events.jsonl[.*]` under the Go section. The default `AO_DATA_DIR` lives outside the repo, so nothing leaks today; this guards against a data dir pointed at the tree during local dev.

## Test plan
- [x] `gofmt -l .` clean
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -race ./...` green
- [x] Contract parity: existing lifecycle/session suites pass against the SQLite store
- [x] Revision CAS, outbox atomicity, CDC ordering + idempotency, rotation/resync, janitor vacuum
- [x] Reaction durability across a simulated restart (no duplicate escalation)
- [x] projects: upsert/get/list/delete round-trip, archive hides from ListProjects while GetProject still resolves
- [x] pr_enrichment: upsert/get/delete round-trip
- [x] Smoke boot: daemon starts, creates `ao.db` + WAL + `session-events.jsonl`, drains cleanly on SIGTERM


harshitsinghbhandari:
Code review from the LCM/SM lane (aa-27). Compat is **green** — port shape, schema columns, aa-7 v3 blockers, pragmas, the durable `ReactionStore` seam all line up with what `backend/internal/lifecycle/*` expects. That's covered separately and won't be re-derived here.

This is the **code-quality** pass on the adapter itself. **Verdict: REQUEST-CHANGES** on two correctness gaps; everything else below is "should fix" or nit. Re-reviewed after the two follow-up commits (`23b8fe4` projects + pr_enrichment, `4ce9044` typed metadata + structured PR facts) — neither addresses the findings below; details on the new work at the end.

---

## Blockers

### B1 — Consumer drops live events across restart-with-rotation

**File:** `backend/internal/cdc/consumer.go` `Start` (~84-106) + `Poll` (~107-149)

On `Start`, `c.prevInfo == nil` and `c.cursor = 0`, so neither rotation branch in `Poll` fires on the first tick. If the JSONL rotated between the last successful poll and a restart, the events in `session-events.jsonl.1` are never read into the live stream — only their net effect resurfaces via `session_snapshot` (and only after a *future* rotation, see S3). For an at-least-once contract this is a hole.

**Repro:** `lastSeq=100`, writer emits 101..600 (spans rotation at 1 MiB), SIGTERM lands exactly at rotation, restart. Consumer reads only seq 601+.

**Fix:** on `Start`, if `c.snapshot != nil`, call `c.resync(ctx)` once before the poll loop. Or gap-probe: read one line, resync iff `e.Seq > c.lastSeq + 1`. The closest existing test (`TestConsumerRestartSkipsDelivered`) doesn't exercise rotation-across-restart.

### B2 — `sendToAgent` rollback corrupts shared tracker state, and now persists it

**File:** `backend/internal/lifecycle/reactions.go` `sendToAgent` (~325-358; failure rollback ~348-356)

`sendToAgent` mutates `tk` (a pointer into `m.trackers`), drops the lock for the messenger `Send`, then on failure re-acquires and decrements `tk.attempts--` / zeroes `tk.firstAttemptAt`. Two concurrent callers for the same `(id, key)` can race so the loser's rollback undoes the winner's successful attempt: in-memory `attempts` undercounts, `firstAttemptAt` is zeroed (escalation anchor lost), and `persistTracker` now writes that corrupted state to SQLite.

Pre-PR this was a transient memory glitch that reset on restart. With `WithReactionStore` write-through, `LoadReactionTrackers` re-hydrates the corruption. The file header at `reactions.go:197-209` already acknowledges the broader concurrency class; the durability seam is the right time to address it.

Companion path: `clearSessionTrackers` deletes the map entry while an in-flight `sendToAgent` still holds the orphaned `tk`; the post-Send rollback then re-`persistTracker`s a tracker for a session whose incident is over.

**Fixes (pick one):** (a) snapshot-and-CAS rollback (only roll back if `tk` hasn't been touched), (b) hold the lock across `messenger.Send` or introduce a per-`(session, key)` lock, (c) pass a generation through `PersistedTracker` and reject stale generations server-side. Worth a `-race` test under `reaction_durability_test.go`.

---

## Should fix (top 5)

- **S1** `queries/cdc.sql` — `DeleteSentOutboxBelow` uses `change_log_seq < ?` but the watermark is `MIN(last_seq)` which is *already acknowledged*. One char fix to `<=`; otherwise a quiet daemon never fully vacuums.
- **S2** `cdc/publisher.go` `Drain` — a permanently-failing outbox row blocks the entire pipeline forever (no DLQ, no attempts threshold). At minimum log `attempts`; better, skip rows past N attempts (or move to a `dead_outbox`).
- **S3** `cdc_wiring.go` snapshot stamps **every emitted event with the same `Seq: maxSeq`**, and collapses N original event types into `session_snapshot`. Any subscriber that dedups by seq (a documented idempotency key) will drop N-1 of N snapshot events. Either give each snapshot event a unique seq or scan `change_log` from `lastSeq+1` instead of `ListAll → sessions`.
- **S4** `storage/sqlite/db.go` — `SetMaxOpenConns(1)` serialises every write across unrelated sessions, contradicting the throughput intent of `keyedMutex` in `lifecycle/manager.go`. Either align the docstrings or commit to multi-conn + `BEGIN IMMEDIATE`.
- **S6** `cdc/broadcast.go` `Publish` — holds `RLock` while invoking subscriber callbacks; a blocking or panicking subscriber takes down the consumer loop. Copy subs under RLock, release, then iterate with `recover()`.

Full list (S5/S7-S10 + 12 nits + test gaps) in the off-PR review doc. Happy to pair on B1 or B2 — both fixes are localised.

---

## Re: the two follow-up commits

- `23b8fe4` (projects + pr_enrichment): `migrations/0002_pr_projects.sql` adds `projects` (soft-delete via `archived_at`) and structured PR fact tables (`pr` + `pr_check` + `pr_comment` with FK CASCADE), both off the canonical CDC path. Good shape, no review-worthy issues.
- `4ce9044` (typed metadata + structured PR facts): `session_metadata` retyped from KV bag to a 1:1 typed table backed by `domain.SessionMetadata`, with `json:"-"` on `SessionRecord.Metadata` enforcing the off-CDC invariant at the type level. Strict improvement — the `Meta*` string constants going away removes a class of silent-drop bugs, and the LCM/SM/reaper threading is self-contained (no compat break on the LCM lane).

So the new work doesn't introduce review-worthy issues of its own; it just doesn't address the existing ones. B1 and B2 still want resolving before more code starts depending on the durable tracker invariant.

---

Nothing in the storage adapter itself (revision CAS, outbox atomicity, sqlc usage, Detecting round-trip, metadata side-channel) is incorrect — the parity test against `fakeStore` is the right shape and passes. Bugs are concentrated in the CDC consumer and the new tracker write-through path.

backend/internal/config/config.go
```diff
@@ -138,3 +148,17 @@ func resolveRunFilePath() (string, error) {
...
+// directory alongside running.json.
+func resolveDataDir() (string, error) {
+	if p, ok := os.LookupEnv("AO_DATA_DIR"); ok && p != "" {
+		return p, nil
+	}
+	dir, err := os.UserConfigDir()
+	if err != nil {
+		return "", fmt.Errorf("resolve state dir: %w", err)
+	}
+	return filepath.Join(dir, "agent-orchestrator", "data"), nil
```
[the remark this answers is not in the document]
AgentWrapper: Good catch in principle on config-dir vs state-dir, but `os.UserStateDir()` does not exist in the Go standard library — `os` only provides `UserConfigDir`, `UserCacheDir`, and `UserHomeDir`. The suggested change doesn't compile, so I've kept `os.UserConfigDir()`. If we want the Linux `~/.local/state` split, it'd need an `XDG_STATE_HOME`-aware resolver as a small follow-up.

AgentWrapper:
@greptile please re-review — the branch has been substantially reworked since the last pass (force-pushed onto the trigger-driven storage/CDC redesign, then the whole lifecycle lane ported onto it).

**How the previous review was addressed:**

| Prior finding | Resolution |
|---|---|
| `main.go:61` (P0) double-wrap `NewStore` | Fixed — `store, err := sqlite.Open(...)` used directly. |
| `cdc_wiring.go:38` (P0) stale outbox API | File rewritten to the `trigger → poller → broadcaster` model. |
| `lifecycle_wiring.go:55` (P0) undefined store methods / `LifecycleStore` | Rewritten — `storeAdapter` bridges `*sqlite.Store` to the new `SessionStore`/`PRWriter` ports; durable reaction-trackers removed. |
| `cdc_e2e_test.go:50` (P0) stale test | Removed; replaced by `wiring_test.go` (store → trigger → change_log → poller → broadcaster, end to end). |
| `cdc_wiring.go:136` (P1) snapshot seq collision | Moot — snapshot-resync was removed in the trigger-driven redesign (clients track their own offset). |
| `cdc/jsonl.go:21` (P2) rotation comment | Moot — the JSONL log was removed; CDC is captured by DB triggers. |
| `config/config.go:163` (P2) `UserConfigDir` → `UserStateDir` | Not applied — `os.UserStateDir` doesn't exist in the Go stdlib (see inline reply). Kept `UserConfigDir`. |

The whole backend now builds and `go test -race ./...` is green. The lifecycle lane (LCM, reactions, session manager, reaper, boot) was ported onto the new model and trimmed from ~6.2k to ~2.8k LOC.

backend/main.go
```diff
@@ -46,11 +47,53 @@ func run() error {
...
+	//     so it's deferred rather than faked. The LCM already exposes the read
+	//     surface (RunningSessions) the SM would wrap.
+	//   - HTTP API routes: httpd.New takes no SM/LCM today; surfacing the store
+	//     over HTTP needs a constructor signature change + handlers, tracked with
+	//     the SM work since the routes call into it.
+	lcStack, err := startLifecycle(ctx, store, log)
+	if err != nil {
+		return fmt.Errorf("start lifecycle: %w", err)
+	}
+	defer lcStack.Stop()
```
illegalcall: This can hang on early `srv.Run` failures. The CDC/reaper goroutines stop only when `ctx` is cancelled, but `stop()` is deferred before these `Stop()` waits, so LIFO defer order waits first and cancels last. If `srv.Run(ctx)` returns without a signal, for example `runfile.Write` fails, `lcStack.Stop()`/`cdcPipe.Stop()` can block forever. Please cancel the context before waiting on those goroutines.

backend/internal/storage/sqlite/migrations/0001_init.sql
```diff
@@ -0,0 +1,213 @@
...
+    INSERT INTO change_log (project_id, session_id, event_type, payload, created_at)
+    VALUES ((SELECT project_id FROM sessions WHERE id = NEW.session_id), NEW.session_id, 'pr_updated',
+        json_object('url', NEW.url, 'session', NEW.session_id, 'state', NEW.pr_state,
+                    'ci', NEW.ci_state, 'review', NEW.review_decision, 'mergeability', NEW.mergeability),
+        NEW.updated_at);
+END;
+-- +goose StatementEnd
+
+-- +goose StatementBegin
+CREATE TRIGGER pr_checks_cdc_insert
```
illegalcall, line 192: `UpsertPRCheck` updates an existing `(pr_url, name, commit_hash)` row on conflict, but CDC only has an `AFTER INSERT` trigger for `pr_checks`. A check moving from `in_progress` to `failed`/`passed` on the same commit updates SQLite without emitting `change_log`, so live clients and offset catch-up miss the status change. Please add an `AFTER UPDATE` trigger and a regression test.

backend/internal/lifecycle/manager.go
```diff
@@ -21,441 +16,241 @@ import (
...
-	// fires no reaction — the transition is discarded.
-	_, err := m.mutate(ctx, id, func(cur domain.CanonicalSessionLifecycle, exists bool) (domain.CanonicalSessionLifecycle, bool, error) {
-		if !exists {
-			// Killing an unknown/already-gone session is a benign race; no-op
-			// rather than fabricating a terminal record for a session we never
-			// knew about.
-			return cur, false, nil
+// writePR upserts the scalar facts, records each check run, and replaces the
+// comment set. PR-table CDC is emitted by the DB triggers.
+func (m *Manager) writePR(ctx context.Context, id domain.SessionID, o ports.PRObservation) error {
```
illegalcall, line 181: This writes one PR observation as multiple independent store calls: scalar PR upsert, each check, then comment replacement. If a later step fails or the context is cancelled, the PR row and its CDC event can already be committed while checks/comments are stale or partial. The observation should be persisted in one storage transaction so the DB state and emitted CDC stay consistent.

backend/internal/storage/sqlite/db.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"database/sql"
+	"embed"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/pressly/goose/v3"
+	_ "modernc.org/sqlite"
```
illegalcall, line 17: The stack PR recommends trying `github.com/ncruces/go-sqlite3/driver` first behind `database/sql`. If we are choosing `modernc.org/sqlite` instead, let’s capture the reason in the stack PR and keep that as the source of truth.

backend/main.go
```diff
@@ -46,11 +47,53 @@ func run() error {
 		return err
 	}
 
+	// Open the durable store and bring up the CDC substrate (outbox publisher,
```
illegalcall: This comment still describes the old outbox/JSONL/janitor flow, while the current implementation uses trigger-driven `change_log` polling. Please update this and the related startup comment below to match the implementation.

illegalcall (commented):
We have 2 PRs that recommend the backend architecture and library choices:

- Architecture / code structure: https://github.com/aoagents/agent-orchestrator/pull/41
- Technical stack / libraries: https://github.com/aoagents/agent-orchestrator/pull/25

Let’s try to follow those.

If you disagree with any part of those recommendations, let’s comment/contribute there and keep those PRs as the source of truth.

Based on that, I recommend these changes here:

- Use the recommended SQLite driver path: `github.com/ncruces/go-sqlite3/driver` first, unless the stack PR changes.
- Keep this PR focused on `internal/storage/sqlite` and CDC persistence.
- Correlate the storage/CDC implementation with the architecture PR’s lifecycle/session boundaries.
- If we want to change the lifecycle architecture, let’s discuss that on the architecture PR first.
- Keep lifecycle writes routed through the LCM contract unless the architecture PR changes.
- Preserve the documented per-session LCM write model unless we agree to change it in the architecture PR.
- Align CDC with the recommended storage/change-log/outbox direction, or update/discuss that in the architecture PR first.
- Update stale comments in `main.go` that still mention the old JSONL/outbox flow.

backend/main.go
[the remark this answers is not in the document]
AgentWrapper: Good catch — real deadlock. Fixed in 0dbd304: the goroutines are now drained explicitly after `srv.Run` returns — `stop()` cancels the context first, then `lcStack.Stop()` / `cdcPipe.Stop()` wait — instead of via LIFO defers. Verified by booting the daemon and sending SIGTERM (drains and exits cleanly in <1s); the non-signal path (e.g. listener bind error) now follows the same cancel-then-drain ordering.

backend/main.go
AgentWrapper: Updated in 0dbd304 — both the store-open and lifecycle startup comments now describe the trigger-driven flow (DB trigger → change_log → poller → broadcaster), with no mention of the removed outbox/JSONL/janitor. Thanks!

backend/main.go
AgentWrapper: Fixed in 0dbd304 — same root cause you and Greptile both flagged. The goroutines are now drained explicitly after `srv.Run` returns: `stop()` cancels the context first, then `lcStack.Stop()`/`cdcPipe.Stop()` wait. Verified by booting + SIGTERM (drains and exits cleanly in <1s).

backend/internal/storage/sqlite/migrations/0001_init.sql
AgentWrapper: Great catch — this was the sharpest comment on the PR. Fixed in 70aab5e: added a `pr_checks_cdc_update` trigger guarded on `OLD.status <> NEW.status`, so a status flip on the same commit emits `change_log` while a no-op re-poll stays silent. Added a regression test asserting both (emit on change, suppress no-op).

backend/internal/lifecycle/manager.go
AgentWrapper: Fixed in 70aab5e. `writePR` now persists the scalar facts, checks, and comments in a single transaction via `Store.WritePRObservation` (scalar upsert first so the check/comment CDC triggers can resolve the session id within the same tx). Collapsed the three `PRWriter` write methods into one `WritePR` since the engine only ever writes an observation as a whole. Round-trip test added.

backend/internal/storage/sqlite/db.go
AgentWrapper: Recorded the rationale at the import site in 70aab5e: `modernc.org/sqlite` is the pure-Go / CGO-free driver, chosen so the daemon cross-compiles and ships as a static binary with no libsqlite/CGO toolchain dependency (tradeoff: some raw throughput vs a C-backed driver). Happy to mirror that into the stack PR as the source of truth if you point me at it.

illegalcall (approved):

PR #47, opened 2026-05-30
feat(api): implement project routes with mock manager/store

Vaibhaav-Tiwari:
## Summary

  Implements the Phase 3 `/api/v1/projects` API surface
  on top of PR #24 route shell by replacing `501 Not
  Implemented` stubs with working handlers backed by a
  mock in-memory project manager/store.

  ## What Changed

  - Implemented project route handlers in `backend/
  internal/httpd/controllers/projects.go`:
    - `GET /api/v1/projects`
    - `POST /api/v1/projects`
    - `GET /api/v1/projects/{id}`
    - `PATCH /api/v1/projects/{id}`
    - `DELETE /api/v1/projects/{id}`
    - `POST /api/v1/projects/{id}/repair`
    - `POST /api/v1/projects/reload`
  - Added mock project service layer:
    - `backend/internal/project/manager.go`
    - `backend/internal/project/memory_store.go`
    - `backend/internal/project/errors.go`
  - Wired default mock manager in `backend/internal/
  httpd/api.go` when project deps are not provided.
  - Replaced route-shell 501 tests with behavior tests
  in `backend/internal/httpd/controllers/
  projects_test.go`.
  - Fixed Windows compile issue in `backend/internal/
  runfile/rename_windows.go` (`MoveFileExW` via
  `kernel32.dll`) so backend tests run on Windows.

  ## Behavior / Validation

  - Returns structured API errors (`INVALID_JSON`,
  `PATH_REQUIRED`, `NOT_A_GIT_REPO`,
  `INVALID_PROJECT_ID`, `PROJECT_NOT_FOUND`,
  `IDENTITY_FROZEN`, `PATH_ALREADY_REGISTERED`,
  `ID_ALREADY_REGISTERED`, `REPAIR_NOT_AVAILABLE`).
  - Enforces identity freeze for PATCH on: `projectId`,
  `path`, `repo`, `defaultBranch`.
  - Mock DB is in-memory only (no durable persistence
  yet).

  ## Testing

  Ran locally:

  ```bash
  cd backend
  go test ./...
  go vet ./...

  All passing.

  ## Scope Notes

  - This PR intentionally uses mocked project storage
    as assigned for Phase 3.

  - Follow-up work will wire real persistence (SQLite
    store) and implement real repair/reload/delete side
    effects (session/workspace cleanup).


Vaibhaav-Tiwari:
@copilot resolve the merge conflicts in this pull request

backend/internal/httpd/api.go
```diff
@@ -0,0 +1,98 @@
...
+type API struct {
+	cfg      config.Config
+	projects *controllers.ProjectsController
+}
+
+// NewAPI constructs the API surface from its dependencies. cfg carries the
+// per-request timeout so the REST group can apply it without re-reading the
+// environment.
+func NewAPI(cfg config.Config, deps APIDeps) *API {
+	if deps.Projects == nil {
```
illegalcall: Defaulting nil deps to `project.NewMemoryManager()` makes the production router expose a successful but process-local project API whenever callers forget to wire a real manager. That makes `POST /projects` look durable even though the data disappears on restart. I think the mock manager should be injected explicitly by tests/dev wiring, or nil should keep the 501 behavior until a real project manager is available.

backend/internal/project/manager.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+	row, ok, err := m.store.Get(ctx, string(id))
+	if err != nil {
+		return Project{}, internal("PROJECT_LOAD_FAILED", "Failed to load project")
+	}
+	if !ok {
+		return Project{}, notFound("PROJECT_NOT_FOUND", "Unknown project")
+	}
+
+	// The PR #37 schema stores only registry identity columns. Behaviour config
```
illegalcall: This returns 200 for PATCH while deliberately dropping every requested config field. That is worse than a stub because clients will believe `agent`, `runtime`, `tracker`, etc. were accepted. For the mock phase, can we either persist these fields in the in-memory row/read-model, or return a clear unsupported/not-implemented error until a config store exists?

backend/internal/project/manager.go
```diff
@@ -0,0 +1,251 @@
...
+		})
+	}
+	return out, nil
+}
+
+func (m *manager) Get(ctx context.Context, id domain.ProjectID) (GetResult, error) {
+	if err := validateProjectID(id); err != nil {
+		return GetResult{}, err
+	}
+	row, ok, err := m.store.Get(ctx, string(id))
```
illegalcall, line 53: Because `Remove` archives the row and `Get` does not check `ArchivedAt`, a project remains readable as `status: ok` after DELETE. That conflicts with the route contract’s “remove/unregister” semantics. Please either treat archived rows as not found/degraded at the API layer, or update the architecture/spec discussion to say deleted projects remain fetchable.

backend/internal/project/manager.go
```diff
@@ -0,0 +1,251 @@
...
+		}
+	}
+	abs, err := filepath.Abs(raw)
+	if err != nil {
+		return "", badRequest("INVALID_PATH", "Repository path is invalid", nil)
+	}
+	return filepath.Clean(abs), nil
+}
+
+func isGitRepo(path string) bool {
```
illegalcall, line 216: The stack PR recommends using the real `git` CLI for Git behavior. A filesystem `.git` check accepts fake repos and misses some edge cases that `git -C <path> rev-parse --is-inside-work-tree` would handle. If this endpoint is doing real validation, it should go through the Git CLI path; otherwise keep this explicitly mock-only.

backend/internal/httpd/apispec/openapi.yaml
```diff
@@ -0,0 +1,444 @@
...
+    ReloadResult:
+      type: object
+      required: [reloaded, projectCount, degradedCount]
+      properties:
+        reloaded: { type: boolean }
+        projectCount: { type: integer }
+        degradedCount: { type: integer }
+
+    # ---- Behaviour config blobs (ported from the TS Zod schemas) ----
+    # All carry .passthrough() semantics — extra keys are preserved on
```
illegalcall: This spec says extra config keys are preserved and re-emitted, but the Go DTOs do not carry an `Extra`/raw field and `UpdateConfig` currently drops the whole patch. Please align the spec with the implementation, or implement passthrough preservation before advertising it as part of the contract.

illegalcall (commented):
We have 2 PRs that recommend the backend architecture and library choices:

- Architecture / code structure: https://github.com/aoagents/agent-orchestrator/pull/41
- Technical stack / libraries: https://github.com/aoagents/agent-orchestrator/pull/25

Let’s try to follow those.

If you disagree with any part of those recommendations, let’s comment/contribute there and keep those PRs as the source of truth.

Based on that, I recommend these changes here:

- Keep the `internal/project` feature-package direction, but make sure the mock store does not become the production source of truth by accident.
- Keep HTTP as the protocol adapter: controllers should call `project.Manager`, and storage/config/runtime side effects should stay behind the project feature layer.
- If project registration validates a real git repo, use the recommended Git CLI path instead of filesystem-only `.git` checks.
- Make PATCH/DELETE semantics honest: either persist the requested mock state consistently, or return a clear not-implemented/unsupported response until the real project store lands.
- Align the OpenAPI spec with actual behavior, especially around config passthrough and archived/deleted project visibility.

backend/internal/project/manager.go
```diff
@@ -0,0 +1,255 @@
...
+	cmd := exec.Command("git", "-C", path, "rev-parse", "--show-toplevel")
+	out, err := cmd.Output()
+	if err != nil {
+		return false
+	}
+	top := filepath.Clean(strings.TrimSpace(string(out)))
```
illegalcall, line 222: This equality check rejects valid repos on macOS when the input path is under `/var/...`: `git rev-parse --show-toplevel` returns the canonical `/private/var/...` path, while `normalizePath` keeps `/var/...`, so the same repo fails this comparison. That is why the latest `git init` fixtures fail with `NOT_A_GIT_REPO`. Please canonicalize both paths with `filepath.EvalSymlinks` before comparing, or treat a successful `git -C path rev-parse --show-toplevel` as sufficient for repo validation if nested worktrees are acceptable.

illegalcall (commented):
Rechecked latest head `d46bf77`. The earlier API semantics comments are partly addressed, but `go test ./...` now fails in `backend/internal/httpd/controllers` because valid `git init` fixtures are rejected as `NOT_A_GIT_REPO` on macOS.

Checked:
- `gofmt -l .` passes
- `go vet ./...` passes
- `go test ./...` fails in controllers

backend/internal/project/manager.go
Vaibhaav-Tiwari: i'll canonicalize both paths with filepath.EvalSymlinks before comparing.

Vaibhaav-Tiwari:
@copilot resolve the merge conflicts in this pull request

backend/internal/project/manager.go
```diff
@@ -0,0 +1,264 @@
...
+	path = filepath.Clean(path)
+	top, err = filepath.EvalSymlinks(top)
+    if err != nil {
```
illegalcall: This new block needs `gofmt`; `gofmt -l .` currently reports `backend/internal/project/manager.go`.

illegalcall (commented):
Rechecked latest head `0b0b136`. The `/var` vs `/private/var` git repo fixture issue is fixed now: `go test ./...` and `go vet ./...` pass locally.

Still not ready to merge:

- GitHub currently marks this PR as `CONFLICTING` / `DIRTY` against latest `main` after PR #37 merged. Please rebase/merge latest `main` and resolve the `backend/go.mod` / `backend/go.sum` conflict.
- `gofmt -l .` returns `backend/internal/project/manager.go`.

After that, rerun the backend checks on the rebased branch.

backend/internal/project/manager.go
Vaibhaav-Tiwari: yep working on that only

illegalcall (commented):
Rechecked latest head `5046ea8`. The previous merge conflict and `gofmt` issues are resolved.

Checked locally:
- `gofmt -l .` passes
- `go vet ./...` passes
- `go test ./...` passes
- `go test -race ./...` passes

GitHub now reports the PR as `MERGEABLE` / `CLEAN`.

illegalcall (approved):

PR #50, opened 2026-05-31
feat(backend): PTY-attach terminal streaming over /mux WebSocket

### LINKED ISSUE #51: Port terminal streaming to the Go backend over a /mux WebSocket
## Problem

The legacy TypeScript stack streamed terminal output to the dashboard through a separate Node WebSocket process, because Next.js cannot hijack the HTTP connection to perform a WebSocket upgrade. The Go backend rewrite has no terminal surface yet, so the dashboard cannot attach to a running agent's tmux pane.

## Goal

Provide a single multiplexed WebSocket endpoint on the Go daemon that streams PTY output from a tmux pane to the browser and forwards live session-state events, replacing the legacy out-of-process Node server.

## Requirements

- A `/mux` WebSocket endpoint owned by the daemon, with the upgrade handled in-process by net/http.
- A channel-tagged JSON protocol covering terminal output, subscribe, session-state, and system (ping/pong) channels.
- PTY attach to the tmux pane via exact-match target, with bounded re-attach and backoff so transient drops do not kill the stream.
- A ring buffer that replays recent output to a client on subscribe so a reconnecting terminal is not blank.
- A heartbeat to detect dead connections.
- Session-state updates fed from the CDC broadcaster. Raw PTY bytes must never flow through the CDC change_log.
- The stream logic must live in a dedicated feature package. httpd owns only the upgrade and decoding, and must not reach into the terminal internals.
- Hermetic tests via an injectable spawn seam, plus a tmux-guarded integration test, all green under the race detector.

## Notes

This is a straight port of the legacy behavior to the Go architecture. The transport library is coder/websocket and the PTY is creack/pty on Unix. Windows returns a clear not-supported error for now.

Pritom14:
Closes #51

## Summary

Ports the terminal-streaming surface from the legacy TypeScript repo to the Go backend. In the old stack a separate Node WebSocket process was required because Next.js cannot hijack the HTTP connection for an upgrade. Go's net/http upgrades in-process, so this lands as a single multiplexed `/mux` endpoint owned by the daemon, with the stream logic in a dedicated feature package.

The change is split into three reviewable commits: the feature package with its tests, the httpd transport that mounts it, and the daemon wiring.

## What changed

**New `internal/terminal` feature package (owns the workflow):**
- `protocol.go`: the channel-tagged JSON wire protocol. Channels are `terminal`, `subscribe`, `sessions`, and `system`. Terminal payloads are base64 because raw PTY bytes are not guaranteed valid UTF-8.
- `session.go`: a single attached PTY session. Defines the `PTYSource` seam (supplies the attach command and liveness) and the `ptyProcess`/`spawnFunc` seam that lets fan-out, ring replay, and re-attach be tested without a real PTY, tmux, or network. Includes bounded re-attach with backoff for resilience across transient tmux drops.
- `ring.go`: a 50KB ring buffer that replays recent output to a client on subscribe so a reconnecting terminal is not blank.
- `manager.go`: the `Manager` that accepts connections, multiplexes channels, runs the write loop and heartbeat, and bridges the CDC broadcaster onto the `sessions` channel. Defines its `EventSource` and `wsConn` interfaces next to where they are consumed.
- `pty_unix.go` / `pty_windows.go`: the real spawn via creack/pty on Unix; a clear not-supported error on Windows for now.

**`internal/httpd` transport (owns only upgrade and decoding):**
- `mux.go`: mounts `/mux`, performs the WebSocket upgrade, sets a 1MiB read limit, and adapts coder/websocket to the manager's `wsConn`. It does not reach into the terminal internals. When the manager is nil the route is simply not mounted.
- `server.go` / `router.go`: thread the optional `*terminal.Manager` through `New` and `NewRouter`.

**Daemon wiring (`main.go`):**
- Constructs the tmux runtime adapter as the `PTYSource`, builds the terminal manager fed by the CDC broadcaster, and passes it into `httpd.New`. Raw PTY bytes never flow through the CDC `change_log`; only session-state events reach the `sessions` channel via the broadcaster.

## How the flow looks

```
browser xterm.js
      |  WebSocket /mux  (channel-tagged JSON frames)
      v
httpd.muxHandler  (upgrade + 1MiB read limit + decode only)
      v
terminal.Manager.Serve  (multiplex channels, write loop, heartbeat)
      |                         \
      | terminal channel         \ sessions channel
      v                           v
terminal.session              cdc.Broadcaster.Subscribe
  PTYSource.AttachCommand       (session-state events only)
  tmux attach -t =<id>
      v
creack/pty  <-->  ring buffer (50KB, replayed on subscribe)
```

A client opens a terminal channel; the session attaches to the tmux pane, streams PTY bytes out base64-encoded, and replays the ring buffer to late subscribers. In parallel the `sessions` channel forwards CDC session-state events from the broadcaster so the UI can refresh without polling. If the pane drops, the session re-attaches with bounded backoff; on clean exit it reports `exited` and stops.

## Testing

- The `spawnFunc` seam keeps the bulk of the suite hermetic: ring replay, fan-out, re-attach, protocol wire shape, and manager multiplexing all run without a real PTY or network.
- `mux_test.go` exercises the genuine upgrade plus wsjson plus Serve plus creack/pty path against a throwaway shell.
- `session_integration_test.go` drives a real tmux pane (`TestSessionStreamsRealTmuxPane`), guarded so it skips where tmux is absent.
- Full suite is green under `go test -race ./...`; `go build ./...` and `go vet ./...` are clean.

## Notes

This was rebased onto the current main, which carries the trigger-driven CDC refactor. The `sessions` channel payload was updated to match the new `cdc.Event` shape (typed `Type` enum, `ProjectID` added, `revision` dropped).


backend/internal/terminal/session.go
```diff
@@ -0,0 +1,307 @@
...
+	s.subs[id] = onData
+	if onExit != nil {
+		s.exitSubs[id] = onExit
+	}
+	replay := s.ring.snapshot()
+	s.mu.Unlock()
+
+	if len(replay) > 0 {
+		onData(replay)
+	}
```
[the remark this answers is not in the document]
Pritom14: Good catch, fixed in cd84d94. The race is real, but moving only the replay under the lock would not have closed it on its own.

The duplicate has two independent halves. Replay was delivered after the snapshot's lock was released, and copyOut did ring.append and fanout as two separate lock acquisitions. So a chunk appended just before the snapshot could be fanned out just after the replay, and the subscriber received it in both the snapshot and the fanout. Delivering the replay under the lock alone does not help here, because the append already happened before the snapshot while the fanout still lands after the replay returns.

The complete fix needs both critical sections to be serialized:

1. append + fanout are now one atomic step under s.mu (a new deliver method that copyOut calls).
2. the snapshot is replayed before subscribe releases s.mu.

With both holding the same lock, the two sections can never interleave. A chunk is either already in the snapshot, in which case its fanout ran before this subscriber registered, or it is fanned out after subscribe returns, never both. That also fixes the ordering variant you noted, where an exited frame could overtake the replay. onData is a non-blocking enqueue, so holding s.mu across delivery cannot deadlock.

Note that go test -race did not flag this: every access is already lock-protected, so it is a logical ordering bug rather than a memory data race.

backend/internal/terminal/ring_test.go
```diff
@@ -0,0 +1,38 @@
+package terminal
+
+import (
+	"bytes"
+	"strings"
+	"testing"
```
AgentWrapper, line 6: can we make sure all test files are wired into github CI checks?

backend/internal/terminal/session_integration_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"testing"
+	"time"
+
+	"github.com/aoagents/agent-orchestrator/backend/internal/adapters/runtime/tmux"
+	"github.com/aoagents/agent-orchestrator/backend/internal/ports"
+)
+
+// TestSessionStreamsRealTmuxPane attaches a real PTY to a real tmux session and
+// asserts output streams back, then that killing the pane stops the session
+// without a re-attach storm. Skipped when tmux is unavailable.
```
AgentWrapper, line 16: can we make sure all test files are wired into github CI checks?

backend/internal/terminal/protocol_test.go
```diff
@@ -0,0 +1,59 @@
+package terminal
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"testing"
+)
+
```
AgentWrapper, line 8: same here

backend/internal/terminal/fakes_test.go
```diff
@@ -0,0 +1,171 @@
+package terminal
```
AgentWrapper, line 1: what is the purpose of this test? is this needed?

AgentWrapper (approved):

backend/internal/terminal/manager.go
```diff
@@ -0,0 +1,367 @@
...
+			})
+		},
+		func() {
+			c.enqueue(serverMsg{Ch: chTerminal, ID: id, Type: msgExited})
+		},
+	)
+	c.mu.Lock()
+	c.terms[id] = unsub
+	c.mu.Unlock()
+	c.enqueue(serverMsg{Ch: chTerminal, ID: id, Type: msgOpened})
```
illegalcall: A concrete user path for this race: the user opens/reopens the AO terminal for a session that just finished or crashed. `markExited` has already fired, but the manager goroutine has not deleted the session from `m.sessions` yet, so `s.subscribe` immediately enqueues `exited`. We then still store the noop unsubscribe and send `opened`, leaving `c.terms[id]` set; future `open` frames for the same terminal on this socket are ignored until close/reconnect. Can we make the already-exited subscribe path skip storing/sending `opened`, or have the exit callback clear this connection entry?

Pritom14: addressed
`backend/internal/terminal/manager_test.go` - has the tests covered

illegalcall (approved):

PR #59, opened 2026-05-31
feat(frontend): typed API client generated from the OpenAPI spec + gen-verify CI

AgentWrapper:
## Why (scope change)

This PR originally added the code-first OpenAPI generator + a frontend client. The **backend generator has since landed on `main`** (via #68/#65) and now covers projects **and** sessions — superseding this branch's backend work. Rather than duplicate it, this PR is repurposed to the parts `main` still lacks.

## What's here

- **`frontend/src/api/schema.d.ts`** — generated from `backend/internal/httpd/apispec/openapi.yaml` via `openapi-typescript` (`npm run gen:api`). Covers projects, sessions, and orchestrators.
- **`frontend/src/api/client.ts`** — a small `openapi-fetch` client typed by that schema, so the renderer's request/response types come from the daemon contract instead of being hand-maintained.
- **`gen-verify` CI job** — regenerates the spec from Go *and* the TS client from the spec, failing if either committed artifact is stale. Backend drift is already covered by the `apispec` tests; this additionally guards the **frontend** artifact, which nothing else checks.

## Verification

```
frontend: npm run gen:api ✓   npm run typecheck ✓
pipeline: go generate ./... + npm run gen:api reproduce the committed files with no drift
```

Rebased on current `main`; no overlap with the backend codegen or the recent CLI PRs.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -7,6 +7,9 @@
     "": {
       "name": "agent-orchestrator-frontend",
       "version": "0.0.0",
+      "dependencies": {
+        "openapi-fetch": "^0.17.0"
+      },
       "devDependencies": {
         "electron": "^33.0.0",
         "typescript": "^5.6.0"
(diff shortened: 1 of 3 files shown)


neversettle17-101 (approved):
Changes LGTM

PR #68, opened 2026-06-01
refactor: move project manager to service layer

neversettle17-101:
## Summary

Rebases the project-manager cleanup onto latest `main` and aligns service code under resource-specific service packages.

- removes the process-local project `MemoryStore` / `NewMemoryManager`
- moves controller-facing project business logic from `backend/internal/project` into `backend/internal/service/project`
- moves session service code into `backend/internal/service/session`
- moves PR observation manager into `backend/internal/service/pr`
- deletes the old `backend/internal/project` package entirely
- adds `domain.ProjectRecord` as the durable storage row shape shared by service and sqlite without creating service↔storage cycles
- updates sqlite project storage to expose explicit project persistence methods (`ListProjects`, `GetProject`, `FindProjectByPath`, `UpsertProject`, `ArchiveProject`)
- keeps HTTP controllers depending on resource service contracts, not concrete storage
- trims project routes to the currently implemented surface: list, add, get, delete
- preserves the session routes merged in #67 while extending the code-first OpenAPI generator to cover both project and session routes
- regenerates `openapi.yaml`

## Notes

The old import-cycle concern is handled by putting the persistence row in `domain` and the service-facing store interface in `service/project`; sqlite implements that interface implicitly. `service/project` does not import sqlite, and sqlite does not import `service/project`.

## Testing

- `npm run lint`


Vaibhaav-Tiwari:
go test ./... is currently failing on PR 68 in backend/internal/httpd/apispec/specgen: the drift test says the embedded backend/internal/httpd/apispec/openapi.yaml is stale compared with fresh specgen.Build() output. I confirmed that running go generate ./... from the backend/ directory regenerates the spec and makes go test ./internal/httpd/apispec/specgen pass, so the fix is to regenerate and commit openapi.yaml. Since this surfaced on a Windows checkout, it may also be worth adding a .gitattributes rule such as backend/internal/httpd/apispec/openapi.yaml text eol=lf, or normalizing line endings in the test, so the byte-for-byte drift check does not fail across Windows/Linux checkouts

AgentWrapper (approved):

PR #76, opened 2026-06-02
feat: add provider-neutral SCM observer

### LINKED ISSUE #75: Plan: provider-neutral SCM observer
# SCM Observer Plan

## Goal

Build a provider-neutral SCM observer that is injected into the daemon, configured with a GitHub provider for v1, and decouples SCM provider details from the lifecycle manager.

The observer owns polling, provider interaction, cache checks, semantic diffing, database persistence, and lifecycle notification. The lifecycle manager receives only a new provider-neutral SCM DTO, not provider-specific GitHub data and not the older `ports.PRObservation` as the primary contract.

## Settled Decisions

```text
- CI log tail length is fixed at the last 20 lines.
- Lifecycle receives new ports.SCMObservation through ApplySCMObservation.
- ports.PRObservation is not the primary observer-to-LCM contract.
```

## Runtime Model

The daemon wires the observer once at startup:

```text
daemon
  -> configure GitHub provider/client
  -> configure SCM observer
  -> observer.Start(ctx)
```

`observer.Start(ctx)` starts one goroutine:

```text
observer goroutine
  -> run initial Poll() immediately in the background
  -> create 30s ticker
  -> run Poll() synchronously on every tick
  -> exit when ctx is cancelled
```

The initial poll should not block daemon startup. It runs asynchronously inside the observer goroutine.

## Poll Algorithm

Each poll runs in this order:

```text
1. List active sessions from DB.
2. Derive SCM subjects from sessions + projects.
3. Split subjects:
     - sessions with known PR
     - sessions without known PR
4. For each repo, run repo PR-list ETag guard.
5. For sessions without PR:
     - if repo PR-list changed or has no cached ETag, detect PR by branch
     - if repo PR-list returned 304, skip detectPR for that repo this poll
6. Merge newly detected PRs into the known PR set.
7. For known PRs:
     - repo PR-list ETag marks metadata-refresh candidates
     - commit check-runs ETag marks CI-refresh candidates
8. Union refresh candidates:
     - newly detected PRs
     - repo metadata changed PRs
     - CI changed PRs
     - PRs missing enough local state
9. GraphQL batch candidates in chunks of 25.
10. If CI is failing and failed-check fingerprint changed, fetch failed job log tail separately.
11. Every 2 minutes, fetch review threads for PRs that need review refresh.
12. Compute mergeability from current normalized PR/CI/review state.
13. Compare semantic hashes against DB state.
14. Persist changed state to DB.
15. After successful DB write, call lifecycle manager for meaningful changes.
```

## Provider Responsibilities

The GitHub provider owns GitHub-specific behavior:

```text
- token/client setup
- repo PR-list ETag guard
- detect PR by branch
- commit check-runs ETag guard
- GraphQL PR batch fetch
- failed CI log-tail fetch
- review thread fetch
- GitHub error/rate-limit normalization
```

The provider must not call the lifecycle manager directly.

Provider implementations return normalized SCM data. They should not expose raw GitHub payloads to the observer or lifecycle manager.

## Observer Responsibilities

The observer owns orchestration:

```text
- polling loop
- session/project discovery
- in-memory provider cache maps
- refresh candidate selection
- semantic hash comparison
- database persistence
- lifecycle notification after successful persistence
```

The observer should call lifecycle through a new SCM-specific port method, for example:

```go
type LifecycleManager interface {
    ApplySCMObservation(ctx context.Context, sessionID domain.SessionID, obs ports.SCMObservation) error
}
```

The observer should not directly build or send `ports.PRObservation`. If the existing lifecycle internals still need `PRObservation`, the lifecycle package can project `SCMObservation` into `PRObservation` internally.

## Lifecycle DTO

Add a new provider-neutral DTO for observer-to-LCM communication:

```go
type SCMObservation struct {
    Fetched    bool
    ObservedAt time.Time

    Provider string
    Host     string
    Repo     string

    PR SCMPRObservation
    CI SCMCIObservation
    Review SCMReviewObservation
    Mergeability SCMMergeabilityObservation

    Changed SCMChanged
}

type SCMChanged struct {
    Metadata bool
    CI       bool
    Review   bool
}

type SCMPRObservation struct {
    URL          string
    Number       int
    State        string
    Draft        bool
    Merged       bool
    Closed       bool
    SourceBranch string
    TargetBranch string
    HeadSHA      string
    Title        string
    Additions    int
    Deletions    int
}

type SCMCIObservation struct {
    Summary           string
    HeadSHA           string
    FailedFingerprint string
    FailedChecks      []SCMCheckObservation
    FailureLogTail    string
}

type SCMCheckObservation struct {
    Name       string
    Status     string
    Conclusion string
    URL        string
    LogTail    string
}

type SCMReviewObservation struct {
    Decision string
    Threads  []SCMReviewThreadObservation
}

type SCMReviewThreadObservation struct {
    ID       string
    Path     string
    Line     int
    Resolved bool
    IsBot    bool
    Comments []SCMReviewCommentObservation
}

type SCMReviewCommentObservation struct {
    ID       string
    Author   string
    Body     string
    URL      string
    IsBot    bool
}

type SCMMergeabilityObservation struct {
    State       string
    Mergeable   bool
    Conflict    bool
    BehindBase  bool
    Blockers    []string
}
```

Field names can be adjusted during implementation, but this is the intended boundary:

```text
SCM observer -> ports.SCMObservation -> lifecycle.ApplySCMObservation
```

`ports.PRObservation` remains an internal compatibility type for the existing PR reaction path. It should not be the new observer/provider boundary.

### Repo PR-List ETag

Cheap repo-level guard:

```http
GET /repos/{owner}/{repo}/pulls?state=open&sort=updated&direction=desc&per_page=1
If-None-Match: <etag>
```

Behavior:

```text
304:
  open PR list for the repo is unchanged
  skip detectPR for PR-less sessions in that repo
  skip repo-metadata-driven refresh

200:
  update ETag
  repo may have PR metadata or branch changes
  run detectPR for PR-less sessions
  include known PRs as metadata refresh candidates
```

CI can change without the repo PR-list ETag changing, so CI gets its own guard.

### Commit Check-Runs ETag

Per commit guard:

```http
GET /repos/{owner}/{repo}/commits/{head_sha}/check-runs?per_page=1
If-None-Match: <etag>
```

Behavior:

```text
304:
  check-runs for this commit are unchanged

200:
  update ETag
  mark PR as CI refresh candidate
```

## In-Memory Provider Cache

Provider-specific cache stays in memory for v1. It is not persisted in DB.

```go
type ObserverCache struct {
    RepoPRListETag   map[RepoKey]string
    CommitChecksETag map[CommitKey]string
    ReviewETag       map[PRKey]string
    LastReviewPollAt map[PRKey]time.Time
}
```

Cold daemon restart loses these ETags and causes fresh provider calls. That is acceptable for v1.

The maps should be bounded with LRU or simple max-size eviction.

## GraphQL Batch

Batch PR fetches in chunks of 25.

The batch query should fetch PR/check metadata:

```text
- title
- state
- isDraft
- mergeable
- mergeStateStatus
- reviewDecision
- headRefName
- headRefOid
- baseRefName
- additions/deletions
- statusCheckRollup
- CheckRun name/status/conclusion/detailsUrl/databaseId
- StatusContext context/state/targetUrl
- pageInfo.hasNextPage
```

The GraphQL batch does not fetch CI log bodies.

## CI Log Fetch

CI logs are fetched separately, only for failed checks.

Use direct GitHub REST where possible:

```http
GET /repos/{owner}/{repo}/actions/jobs/{job_id}/logs
```

Fetch log tail only when:

```text
- CI is failing
- failed-check fingerprint changed, or no previous log tail exists
```

Store only the last 20 lines.

## Review Threads

Review-thread polling is separate from the 30s PR/CI path.

Run review refresh every 2 minutes when:

```text
- reviewDecision == changes_requested
- reviewDecision changed this poll
- existing review_hash is non-empty and reviewDecision is no longer changes_requested
- local review data is missing or stale
```

Use cursors for pagination, not as the only correctness mechanism.

Correctness should come from semantic hashing:

```text
fetch review threads
normalize threads/comments
compute review_hash
if review_hash changed:
  persist review state
  call LCM with review facts
```

## Existing Tables

Current DB already has:

```text
projects
sessions
pr
pr_checks
pr_comment
change_log
```

Relevant existing fields:

```text
sessions:
  id, project_id, branch, workspace_path, runtime_handle_id, is_terminated

pr:
  url, session_id, number, pr_state, review_decision, ci_state, mergeability, updated_at

pr_checks:
  pr_url, name, commit_hash, status, url, log_tail, created_at

pr_comment:
  pr_url, comment_id, author, file, line, body, resolved, created_at
```

## Proposed DB Changes

`pr` should hold the full latest PR metadata as explicit provider-neutral columns. Do not keep only a few scalar fields plus hashes; the metadata row should be enough for UI/API/lifecycle reads without unpacking provider payloads.

Extend `pr` with generic SCM metadata fields:

```text
provider TEXT NOT NULL DEFAULT ''
host TEXT NOT NULL DEFAULT ''
repo TEXT NOT NULL DEFAULT ''
source_branch TEXT NOT NULL DEFAULT ''
target_branch TEXT NOT NULL DEFAULT ''
head_sha TEXT NOT NULL DEFAULT ''
title TEXT NOT NULL DEFAULT ''
additions INTEGER NOT NULL DEFAULT 0
deletions INTEGER NOT NULL DEFAULT 0
changed_files INTEGER NOT NULL DEFAULT 0
author TEXT NOT NULL DEFAULT ''
base_sha TEXT NOT NULL DEFAULT ''
merge_commit_sha TEXT NOT NULL DEFAULT ''

is_draft INTEGER NOT NULL DEFAULT 0
is_merged INTEGER NOT NULL DEFAULT 0
is_closed INTEGER NOT NULL DEFAULT 0

provider_state TEXT NOT NULL DEFAULT ''
provider_mergeable TEXT NOT NULL DEFAULT ''
provider_merge_state_status TEXT NOT NULL DEFAULT ''
html_url TEXT NOT NULL DEFAULT ''

created_at_provider TIMESTAMP
updated_at_provider TIMESTAMP
merged_at_provider TIMESTAMP
closed_at_provider TIMESTAMP

metadata_hash TEXT NOT NULL DEFAULT ''
ci_hash TEXT NOT NULL DEFAULT ''
review_hash TEXT NOT NULL DEFAULT ''

observed_at TIMESTAMP
ci_observed_at TIMESTAMP
review_observed_at TIMESTAMP
```

Existing `pr_state`, `review_decision`, `ci_state`, and `mergeability` remain the normalized AO-facing fields. The new provider-neutral metadata columns preserve the complete latest PR metadata needed by product surfaces and lifecycle decisions.

Optionally extend `pr_checks`:

```text
conclusion TEXT NOT NULL DEFAULT ''
details TEXT NOT NULL DEFAULT ''
```

Extend `pr_comment`:

```text
thread_id TEXT NOT NULL DEFAULT ''
url TEXT NOT NULL DEFAULT ''
is_bot INTEGER NOT NULL DEFAULT 0
```

Add review thread table:

```sql
CREATE TABLE pr_review_threads (
    pr_url        TEXT NOT NULL REFERENCES pr (url) ON DELETE CASCADE,
    thread_id     TEXT NOT NULL,
    path          TEXT NOT NULL DEFAULT '',
    line          INTEGER NOT NULL DEFAULT 0,
    resolved      INTEGER NOT NULL DEFAULT 0,
    is_bot        INTEGER NOT NULL DEFAULT 0,
    semantic_hash TEXT NOT NULL DEFAULT '',
    updated_at    TIMESTAMP NOT NULL,
    PRIMARY KEY (pr_url, thread_id)
);
```

Do not add provider-specific ETag/cache tables for v1.

## Lifecycle Notification

Lifecycle should be called only after DB persistence succeeds.

Call lifecycle only when semantic state changed:

```text
metadata_hash changed
ci_hash changed
review_hash changed
```

CI failing notification should include:

```text
- PR URL/number
- head SHA
- CI failing state
- failed checks
- log tails
```

Review notification should include:

```text
- reviewDecision
- pending human comments
- thread/comment metadata
```

Merged/closed/mergeability notification should include:

```text
- PR state
- mergeability
- review decision
- CI state
```

The lifecycle manager should accept `ports.SCMObservation`. Internally it may:

```text
SCMObservation
  -> update canonical lifecycle state
  -> optionally project to existing PR reaction/write helpers
```

This keeps future providers decoupled from the old PR observation DTO.

## Implementation Checklist

1. Add SCM observer package.
2. Add GitHub provider package.
3. Add SQLite migration for PR/review schema changes.
4. Add store methods for SCM reads and transactional writes.
5. Add daemon wiring for GitHub provider + observer start/stop.
6. Add `ports.SCMObservation` and `LifecycleManager.ApplySCMObservation`.
7. Add lifecycle adapter/projection from `SCMObservation` to existing PR write/reaction behavior where useful.
8. Add tests for polling, ETag guards, batching, CI logs, review refresh, DB writes, and lifecycle calls.

## Tests

Required coverage:

```text
- observer starts async and performs immediate initial poll
- poll exits on context cancellation
- repo PR-list ETag 304 skips detectPR
- repo PR-list ETag 200 runs detectPR
- newly detected PR is included in same poll refresh
- CI ETag change triggers GraphQL batch even when repo PR-list ETag is unchanged
- GraphQL batch chunks at 25 PRs
- failing CI fetches log tail only when failed fingerprint changed
- review polling respects 2-minute interval
- review hash drives persistence and lifecycle notification
- unchanged hashes do not call lifecycle
- DB transaction persists metadata, CI, checks, reviews, and comments together
```

whoisasx:
Closes #75

## Summary 
add provider-neutral ports.SCMObservation DTOs and lifecycle ApplySCMObservation projection
- implement SCM observer polling with repo/commit ETag guards, PR detection, GraphQL batching, CI log-tail enrichment, review refresh, semantic hashing, DB writes, and lifecycle notification
- extend SQLite PR/check/review schema and store transactional SCM persistence
- wire daemon to start the GitHub-backed SCM observer when GitHub credentials are available
- ## Tests
- go test ./..
- go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.1.6 run ./...


backend/internal/adapters/scm/github/observer_provider.go
```diff
@@ -0,0 +1,551 @@
...
+  ... on CheckRun { name status conclusion detailsUrl url databaseId }
+  ... on StatusContext { context state targetUrl }
+} pageInfo{ hasNextPage } } } } } }
+`, "CONTEXT_LIMIT", strconv.Itoa(scmBatchCheckContextLimit))
+}
+
+func scmObservationFromGraphQL(ref ports.SCMPRRef, pr map[string]any) ports.SCMObservation {
+	checks := scmChecksFromGraphQL(pr)
+	failed := failedSCMChecks(checks)
+	ci := string(ciSummaryFromGraphQL(pr))
```
illegalcall: Blocking: this uses `ciSummaryFromGraphQL` for the observer summary, but that helper derives from the visible `contexts` page and returns `unknown` when contexts are paginated. GitHub `statusCheckRollup.state` is already the aggregate across all contexts; the TS source used that for CI status and only treated context pagination as a loss of per-check details. With more than 100 contexts, an aggregate `FAILURE` can become `unknown` here, so AO misses `ci_failed` status and repair nudges. Please drive `obs.CI.Summary` from the rollup state and only gate `Checks`/`FailedChecks` completeness on pagination.

backend/internal/adapters/scm/github/observer_provider.go
```diff
@@ -0,0 +1,551 @@
...
+		addBlocker("changes_requested")
+	}
+	if out.State == string(domain.MergeBlocked) {
+		return out
+	}
+	if state == "UNSTABLE" {
+		out.State = string(domain.MergeUnstable)
+		return out
+	}
+	if mergeable == "MERGEABLE" && (state == "CLEAN" || state == "HAS_HOOKS" || state == "") {
```
illegalcall: Blocking: this can mark a PR mergeable while it still has `ReviewRequired` (and the helper also has no draft signal). The status derivation checks `MergeMergeable` before `ReviewRequired`, so a review-pending PR can surface as merge-ready. The source implementation only sets merge-ready for review `approved`/`none` and non-draft PRs. Please include review-required/draft as blockers before returning `MergeMergeable` (or pass those signals into this helper and gate the status there).

backend/internal/observe/scm/observer.go
```diff
@@ -0,0 +1,833 @@
...
+		}
+		if s.known.HeadSHA != "" {
+			commitKey := commitKey(s.repo, s.known.HeadSHA)
+			prev := o.Cache.CommitChecksETag[commitKey]
+			res, err := o.provider.CommitChecksGuard(ctx, s.repo, s.known.HeadSHA, prev)
+			if err != nil {
+				o.logger.Error("scm observer: commit check-runs guard failed", "pr", s.known.URL, "sha", s.known.HeadSHA, "err", err)
+			} else {
+				if res.ETag != "" {
+					o.cacheSetString(o.Cache.CommitChecksETag, &o.Cache.commitOrder, commitKey, res.ETag)
```
illegalcall: Blocking: this advances the commit-check ETag before the expensive fetch and DB write have succeeded. If Guard 2 returns 200 with a new ETag, then `FetchPullRequests` or `WriteSCMObservation` fails, the next poll can get 304 and skip a CI transition that was never persisted. The repo PR-list ETag above has the same issue. Please only commit guard ETags after the corresponding observation is fetched and stored, or track failed refreshes so the next poll forces a retry.

backend/internal/observe/scm/observer.go
```diff
@@ -0,0 +1,833 @@
...
+		if hasObs && obs.Review.Decision != "" {
+			decision = obs.Review.Decision
+		}
+		if !o.needsReviewRefresh(pkey, s.known, decision, hasObs, now) {
+			continue
+		}
+		review, err := o.provider.FetchReviewThreads(ctx, ports.SCMPRRef{Repo: s.repo, Number: s.known.Number, URL: s.known.URL})
+		if err != nil {
+			o.logger.Error("scm observer: review refresh failed", "pr", s.known.URL, "err", err)
+			continue
```
illegalcall, line 700: Blocking: on review-thread fetch failure we continue, but a batch observation from the same poll can still persist a new review decision/hash while `replaceReview=false` preserves old `pr_comment` rows. Display status treats those old unresolved comments as active, and after the decision changes `needsReviewRefresh` can stop retrying. On review fetch failure, please avoid updating the review decision/hash or mark review data stale so we retry without treating preserved comments as current.

backend/internal/observe/scm/observer.go
```diff
@@ -0,0 +1,833 @@
...
+		}
+		if !hasObs {
+			obs = observationFromLocal(s.repo, s.known)
```
illegalcall: Blocking: this review-only fallback synthesizes an observation from local PR facts. That synthetic observation lacks checks, failed fingerprints/log tails, and full mergeability blockers, but later persistence recomputes `metadata_hash`/`ci_hash` and sets `CIObservedAt` from this observation. A review-only refresh can therefore overwrite CI/metadata semantic hashes and claim CI was freshly observed. Please preserve local CI/metadata hashes and timestamps when `!hasObs`, or split this into a review-only write path.

illegalcall (commented):
Detailed SCM observer migration review. I focused on behavior that can change persisted PR/CI/review state or user-visible session status.

illegalcall (commented):
Follow-up after a second pass: two more observer-level failure modes that look worth fixing before merge.

backend/internal/observe/scm/observer.go
```diff
@@ -0,0 +1,833 @@
...
+			continue
+		}
+		if pr.Number <= 0 {
+			continue
+		}
+		s.known = domain.PullRequest{URL: pr.URL, SessionID: s.session.ID, Number: pr.Number, SourceBranch: pr.SourceBranch, TargetBranch: pr.TargetBranch, HeadSHA: pr.HeadSHA, Provider: s.repo.Provider, Host: s.repo.Host, Repo: repoFullName(s.repo), UpdatedAt: now}
+		s.hasPR = true
+		s.newlyDetected = true
+		delete(subjects, oldKey)
+		subjects[prKey(s.repo, pr.Number)] = s
```
illegalcall: Blocking: this map assignment silently overwrites any earlier subject that resolved to the same repo+PR key. If two active sessions are on the same branch, or if duplicate sessions both discover the same PR in this poll, only the last one remains in `subjects`; the other session never gets the PR write/lifecycle notification, and a later write may also trip the store PR-reassignment guard. Please keep a slice/list of subjects per PR key or detect duplicate ownership and leave the duplicate session unassigned with an explicit warning instead of overwriting it.

backend/internal/observe/scm/observer.go
```diff
@@ -0,0 +1,833 @@
...
+		if !prepared.Changed.Metadata && !prepared.Changed.CI && !prepared.Changed.Review {
+			continue
+		}
+		pr, checks, threads, comments := domainFromObservation(subj.session.ID, prepared, local, reviewRefreshed[key], now)
+		if err := o.store.WriteSCMObservation(ctx, pr, checks, threads, comments, reviewRefreshed[key]); err != nil {
+			o.logger.Error("scm observer: DB write failed", "session", subj.session.ID, "pr", pr.URL, "err", err)
+			continue
+		}
+		if o.lifecycle != nil {
+			if err := o.lifecycle.ApplySCMObservation(ctx, subj.session.ID, prepared); err != nil {
```
illegalcall, line 365: Blocking: lifecycle effects are acknowledged only by this callback, but the PR hashes have already been persisted by the time this runs. If `ApplySCMObservation` fails transiently, the next poll can see unchanged semantic hashes and skip the observation entirely, so effects like `MarkTerminated` for merged PRs or CI/review nudges are never retried. Please either retry failed lifecycle notifications independently, store a pending-notification marker, or avoid advancing the semantic hashes until lifecycle has accepted the observation.

backend/internal/observe/scm/observer.go
```diff
@@ -0,0 +1,1129 @@
...
+			preserveLocalMetadata: localOnlyObservations[key],
+			preserveLocalCI:       localOnlyObservations[key],
+			preserveLocalReview:   reviewStale[key],
+		}
+		prepared := o.prepareForPersistence(obs, local, opts, now)
+		if !prepared.Changed.Metadata && !prepared.Changed.CI && !prepared.Changed.Review {
+			prRefreshOK[key] = true
+			continue
+		}
+		if o.lifecycle != nil {
```
illegalcall, line 364: Blocking: this calls lifecycle before the observation is durably written. That is the reverse of the contract above (lifecycle should run after successful persistence), and it creates a one-way split if the DB write fails afterward. The merged-PR path is the worst case: ApplySCMObservation -> ApplyPRObservation calls MarkTerminated, then WriteSCMObservation can fail, and the next poll skips the terminated session so the PR row may never record merged/closed state. CI/review nudges have a similar mismatch via sendOnce: the agent can be nudged and the in-memory reaction marked seen even though the PR/check rows are still stale. Please make the DB write and lifecycle delivery consistent, e.g. persist first with a retryable pending-lifecycle marker/outbox, or otherwise avoid lifecycle side effects until WriteSCMObservation has succeeded.

backend/internal/adapters/scm/github/observer_provider.go
```diff
@@ -0,0 +1,568 @@
...
+	roll := statusRollup(pr)
+	if roll == nil {
+		return domain.CIUnknown
+	}
+	return mapRollupState(str(roll["state"]))
+}
+
+func scmChecksFromGraphQL(pr map[string]any) []ports.SCMCheckObservation {
+	roll := statusRollup(pr)
+	contexts, _ := roll["contexts"].(map[string]any)
```
illegalcall, line 411: Blocking (remaining gap from the CI pagination fix): contexts(first:100) can still be truncated, but scmChecksFromGraphQL ignores contexts.pageInfo.hasNextPage and returns the first page as if it were complete. The latest summary now correctly uses the aggregate rollup state, so a hidden failure produces CI.Summary=failing with FailedChecks=[] and an empty fingerprint. enrichFailureLogs then does not fetch logs, and scmToPRObservation passes no failed check to ApplyPRObservation, so AO can mark the PR as ci_failed without sending the agent the repair nudge/details. The TS source avoided this by leaving ciChecks undefined when hasNextPage was true, forcing a fuller CI fallback. Please either paginate/fetch the full failed context list when rollup is failing and the page is incomplete, or carry an incomplete flag/generic CI-failed nudge so hidden failures are still actionable.

illegalcall (commented):
Follow-up on latest head b3c10b0: prior fixes address most of the earlier findings, but these two current failure modes still look worth fixing before merge.

backend/internal/observe/scm/observer.go
```diff
@@ -0,0 +1,1187 @@
...
+	clock func() time.Time
+	// logger receives non-fatal operational failures.
+	logger *slog.Logger
+	// credentialsChecked records whether an optional provider credential gate ran.
+	credentialsChecked bool
+	// disabled is set after the credential gate reports unavailable credentials.
+	disabled bool
+	// pendingLifecycle retries lifecycle side effects that failed after a
+	// successful DB write, so semantic hash equality cannot make us forget them.
+	pendingLifecycle map[string]pendingLifecycleNotification
```
illegalcall: Blocking: this retry queue is process-local, but the semantic hashes are persisted before the lifecycle side effect is acknowledged. If WriteSCMObservation succeeds, ApplySCMObservation fails, and the daemon restarts before retryPendingLifecycle succeeds, pendingLifecycle is lost. On the next cold poll the PR/check/review hashes already match local state, so prepared.Changed is false and lifecycle will never be retried; merged PR cleanup, CI nudges, or review nudges can be permanently skipped. Please make the pending lifecycle acknowledgement durable (e.g. DB outbox/pending-lifecycle marker consumed by the observer) or avoid advancing the persisted semantic hashes until lifecycle delivery is recorded.

illegalcall (commented):
Follow-up on latest head 481f44a5: the CI pagination fix looks reasonable; one lifecycle retry durability issue remains.

backend/internal/storage/sqlite/migrations/0004_scm_observer_schema.sql
```diff
@@ -0,0 +1,95 @@
...
+    pr_url        TEXT NOT NULL REFERENCES pr (url) ON DELETE CASCADE,
+    thread_id     TEXT NOT NULL,
+    path          TEXT NOT NULL DEFAULT '',
+    line          INTEGER NOT NULL DEFAULT 0,
+    resolved      INTEGER NOT NULL DEFAULT 0,
+    is_bot        INTEGER NOT NULL DEFAULT 0,
+    semantic_hash TEXT NOT NULL DEFAULT '',
+    updated_at    TIMESTAMP NOT NULL,
+    PRIMARY KEY (pr_url, thread_id)
+);
```
Pritom14: `pr_review_threads` has no CDC triggers. Without them, review thread changes never reach `change_log`, so the broadcaster has no visibility into threads being added or resolved, the only live signal consumers get is the rolled-up `review_decision` flip on `pr_updated`, which means there's a window where the agent is resolving individual threads but the dashboard shows no movement until all threads are done.

Two triggers are needed here (add them as separate, `+goose StatementBegin/End` blocks immediately after this table):
 
 1. `AFTER INSERT ON pr_review_threads` → emit `pr_review_thread_added`
 2. `AFTER UPDATE ON pr_review_threads WHEN OLD.resolved <> NEW.resolved` → emit
 `pr_review_thread_resolved`
 
Both need the `change_log.event_type` CHECK constraint updated to include the two new values. SQLitecan't ALTER that constraint in-place -- cleanest fix is to drop and recreate `change_log` in this same migration (safe: it's append-only and the poller re-seeks to head on restart, so no data is lost) or relax the CHECK to `TEXT NOT NULL` and let the trigger layer enforce valid values.

whoisasx:
Implementation plan for the CI/review GraphQL cost tuning discussed in this thread:

### CI contexts

- Reduce the batch GraphQL CI context page from `contexts(first: 100)` to `contexts(first: 20)`.
- Keep `CI.Summary` driven by GitHub's aggregate `statusCheckRollup.state`, not by the visible first page of contexts.
- Add `pageInfo { hasNextPage endCursor }` to the batch query.
- If `hasNextPage=false`, persist the fetched contexts directly.
- If `hasNextPage=true`, run an explicit fallback query for that PR to fetch the remaining CI contexts, append them, and then recompute checks, failed checks, failed fingerprint, and failure details.
- If the CI fallback fails, do not persist partial CI state and do not advance hashes/ETags.

### Review threads

- Use fixed provider constants, not new runtime config flags:
  - `githubReviewThreadPageSize = 50`
  - `githubReviewCommentLimitPerThread = 5`
  - `githubReviewThreadMaxPages = 2`
- Change review fetching to request the latest review window:
  - `reviewThreads(last: 50)`
  - `comments(first: 5)` per thread
  - `pageInfo { hasPreviousPage startCursor }`
- If `hasPreviousPage=false`, treat the review snapshot as complete.
- If `hasPreviousPage=true`, inspect the oldest thread from the latest 50:
  - if the oldest fetched thread is resolved, skip older pages and treat the fetched set as a partial window to merge;
  - if the oldest fetched thread is unresolved, fetch exactly one older page with `reviewThreads(last: 50, before: startCursor)`.
- Never fetch more than two review-thread pages total.

### Persistence behavior

- GitHub provider should not read/write the DB directly. It only returns fetched review data and whether the result is partial.
- Observer chooses the persistence mode:
  - complete review fetch: replace stored review threads/comments;
  - partial review fetch: merge fetched threads/comments while preserving older unseen rows;
  - review fetch failure: preserve existing review rows/hash/decision and retry later.
- For partial merge:
  - upsert fetched review threads;
  - for each fetched thread, replace comments only for that thread;
  - preserve older unseen threads/comments.

### Correctness constraints

- Do not advance CI hash on partial/failed CI fallback.
- Do not treat partial review data as a complete replacement.
- Preserve the existing lifecycle durability behavior: semantic hashes should only advance after lifecycle side effects are accepted.

### Tests to add/update

- CI with <=20 checks does not call fallback.
- CI with >20 checks calls fallback and persists all checks.
- CI summary still follows `statusCheckRollup.state`.
- CI fallback failure does not persist partial CI state.
- Review query shape uses `last:50` and `comments(first:5)`.
- Complete review fetch uses replace mode.
- Partial review fetch uses merge mode.
- Oldest unresolved review thread triggers exactly one older-page fallback.
- Review fetch never exceeds two pages.
- Partial merge preserves older stored rows while updating fetched threads/comments.

whoisasx:
@greptileai

backend/internal/storage/sqlite/migrations/0004_scm_observer_schema.sql
harshitsinghbhandari: Added pr_review_threads_cdc_insert / pr_review_threads_cdc_update triggers and widened change_log.event_type CHECK to include pr_review_thread_added and pr_review_thread_resolved (drop+recreate change_log per your preferred option). cdc.EventType constants added in 0ef2870.

backend/internal/observe/scm/observer.go
```diff
@@ -0,0 +1,1166 @@
...
+			markRepoRefreshFailed(subj.repo)
+			continue
+		}
+		if o.lifecycle != nil {
+			if err := o.lifecycle.ApplySCMObservation(ctx, subj.session.ID, prepared); err != nil {
+				o.logger.Error("scm observer: lifecycle notification failed", "session", subj.session.ID, "pr", firstNonEmpty(prepared.PR.URL, prepared.PR.HTMLURL, local.URL), "err", err)
+				markRepoRefreshFailed(subj.repo)
+				continue
+			}
+			if err := o.store.WriteSCMObservation(ctx, finalPR, finalChecks, finalThreads, finalComments, reviewMode); err != nil {
```
illegalcall: Blocking: the lifecycle acknowledgement write reuses the original reviewMode. When reviewMode is ReviewWriteReplace, WriteSCMObservation deletes all pr_review_threads/pr_comment rows and re-inserts the fetched snapshot. Because this is the second write in the normal successful lifecycle path, every fetched review thread is inserted twice in one poll: once by the held-back-hash write above, then again by this acknowledgement write. The pr_review_threads_cdc_insert trigger emits pr_review_thread_added on each insert, so the broadcaster will see duplicate thread-added events for the same review refresh. The acknowledgement write should only advance the PR semantic hashes/observed timestamps, or at least call WriteSCMObservation with ReviewWritePreserve here so it does not replay review-row replacement after lifecycle has accepted the observation.

illegalcall (commented):
Follow-up on latest head e84d4ee: the previous durability and pagination comments are addressed; I found one new CDC side effect in the acknowledgement write.

harshitsinghbhandari (commented):
High-recall review pass against `feat/75` at `6302844`. 10 findings, ranked by severity — correctness first, then durability/efficiency, then altitude/cleanup. Most prior @illegalcall + @greptile threads look correctly addressed; these are what's left after deduping against resolved threads.

---

### Correctness

**1. `backend/internal/lifecycle/reactions.go:116` — failed-check dedup signature degenerates when `CI.HeadSHA` is empty**

`scmToPRObservation` copies `obs.CI.HeadSHA` into every `PRCheckObservation.CommitHash` (reactions.go:116). On a review-only refresh that fetches threads but no fresh CI batch, the observation flows back through `ApplyPRObservation` with `CommitHash=""`. `ApplyPRObservation` then builds `sig = ch.CommitHash + ":" + ch.LogTail` (reactions.go:52). If both are empty/stable the signature collapses to a constant for that (key, PR, check-name); a later genuinely-new failing CI on a different commit silently no-ops in `sendOnce` because `react.seen[key] == sig` already. The agent never gets the new failure nudge.

**2. `backend/internal/storage/sqlite/queries/pr_comment.sql:5` — `InsertLegacyPRComment` is `INSERT OR IGNORE`, silently dropping updates**

Pre-PR the legacy `WritePR` path replaced the full comment set per PR (DeletePRComments → InsertPRComment). The new `InsertLegacyPRComment` uses `INSERT OR IGNORE` and `writePR` no longer deletes legacy comments first (pr_store.go:74-77 gates delete on `ReviewWriteReplace`). If a reviewer edits the body of an existing comment and the legacy path runs (e.g. through `service/pr.Manager.ObservePR`, today only wired by `integration/lifecycle_sqlite_test.go:97`), the updated body is dropped and the stale body persists. The fix was likely targeted at not clobbering observer-written rows, but it changed the semantics for the legacy path too.

---

### Durability / efficiency

**3. `backend/internal/observe/scm/observer.go:340-374` — two-phase "lifecycle is the cursor" is only durable within one daemon process**

Poll writes the row with preserved (old) hashes (line 358), calls `lifecycle.ApplySCMObservation` (line 364), then writes again with final hashes (line 369). `reactions.sendOnce` dedups in an in-memory `react.seen/attempts` map (reactions.go:172-194). If the daemon crashes after `ApplySCMObservation` returns nil but before the second write commits, the in-memory dedup is lost; on the next start, the same observation re-fetches (hashes unchanged in DB), lifecycle re-fires, and the agent receives a duplicate nudge. The design comment at observer.go:340-345 claims "lifecycle is the cursor" — but the cursor is durable only across polls within the same process. A durable outbox row (or persisting `react.seen/attempts` in the DB) would close this.

**4. `backend/internal/adapters/scm/github/observer_provider.go:168` — `FetchReviewThreads` always fetches a second 50-thread page when the newest thread is unresolved**

After fetching the latest 50 threads, the code issues a second GraphQL page whenever `len(latest) == 0 || !latest[0].Resolved` (observer_provider.go:168). For any actively-reviewed PR the newest thread is typically unresolved, so the 2-minute review cadence fetches 100 threads + their leading 5 comments every cycle, not 50. The intent (recover older unresolved threads) only fires when the newest happens to be unresolved — which is the wrong signal, since older threads need a revisit regardless of the newest one's state. Cost: ~2× GraphQL points per review poll across all open PRs.

**5. `backend/internal/observe/scm/observer.go:339, 357` — `domainFromObservation` is called twice per changed PR; each call recomputes three SHA256 hashes**

When lifecycle is wired, `domainFromObservation` runs once with the pending-hash options (line 338) and again with the final-hash options (line 357). Each call invokes `metadataSemanticHash`, `ciSemanticHash`, and `reviewSemanticHash` (observer.go:765-783). `prepareForPersistence` (line 740) also computes the same three hashes. For 50 changed PRs/poll that's ~450 hash invocations where ~150 suffice. Compute the three hash pairs once in `prepareForPersistence` and thread them through.

**6. `backend/internal/observe/scm/observer.go:626` — `enrichFailureLogs` does O(M×N) match between FailedChecks and all Checks**

For each failed check (~M) the inner loop walks all checks (~N) to copy `LogTail` back into `obs.CI.Checks`. A PR with 80 checks and 10 failing is 800 string comparisons per CI refresh. Pre-build `map[checkKey]int` once, then update `obs.CI.Checks[idx].LogTail` directly. `applyStoredFailedLogTails` (line 635-665) already uses the cleaner `name → tail` map shape — same pattern is applicable here.

---

### Altitude / cleanup

**7. `backend/internal/storage/sqlite/store/pr_store.go:50` — `writePR` mixes legacy and SCM-observer modes via a bool + enum; legacy `WritePR` has no production caller**

`writePR` carries `replaceLegacyComments bool` plus `ReviewWriteMode` and branches on the combination (lines 69-101). Searching the tree, `service/pr.Manager` (the only legacy `WritePR` consumer) is wired only by `integration/lifecycle_sqlite_test.go:97` — no daemon path uses it. Deleting `service/pr.Manager`, `InsertLegacyPRComment`, `genLegacyCommentParams`, and the bool collapses `writePR` to just `WriteSCMObservation`'s body and incidentally resolves finding #2.

**8. `backend/internal/observe/scm/observer.go:1128-1166` — three near-identical FIFO eviction helpers**

`cacheSetString`, `cacheSetTime`, `cacheSetBool`, and `evictStrings` duplicate the same "append-to-order-slice; evict-when-over-cap" pattern, parameterized only by value type. A generic `cacheSet[V any](m map[string]V, order *[]string, key string, value V, max int)` collapses to one helper. Any future fix (proper LRU touch, deterministic eviction, TTL) currently has to land in four places.

**9. `backend/internal/observe/scm/observer.go:716-718` — `ReviewRefreshFailed` reset only flips value to false; `reviewFailedOrder` grows monotonically**

On a successful review refresh, `Cache.ReviewRefreshFailed[pkey] = false` is set but the key is never removed from `reviewFailedOrder` (line 1146-1155 evicts by slice length). Over a long-lived daemon, eviction begins discarding genuinely-failed entries earlier than the configured cap implies, because cleared `false` entries still hold slots. Fix: `delete(map, pkey)` and remove from the order slice on success.

**10. `backend/internal/domain/pr.go:51-53` — provider-specific enum strings leak into the provider-neutral domain**

`domain.PullRequest` carries `ProviderState`, `ProviderMergeable`, `ProviderMergeStateStatus` — raw GitHub strings — alongside the normalized AO enums. `mergeabilityFromProviderFacts` (observer.go:950) inspects GitHub-flavored values like `DIRTY`/`BEHIND`/`UNSTABLE`. A second provider (Linear/GitLab) repeats this contamination. A provider-opaque `ProviderPayload []byte` (or a JSON column) plus moving the provider→AO normalization fully into the provider package keeps the domain layer purely normalized.

---

### Verified refuted (not findings, for transparency)

- `observer_provider.go:406` `scmChecksFromGraphQL` doesn't nil-panic — Go nil-map reads return zero values, and the loop over `nodes(nil["nodes"])` runs zero iterations.
- `reactions.go:122-138` doesn't drop human review comments via the `IsBot` skip — `scmThreadFromGraphQL` (observer_provider.go:555-570) only sets `thread.IsBot=true` when every comment is bot-authored.
- `0004_scm_observer_schema.sql` `change_log` drop+recreate is safe — goose migrations run in a transaction; no concurrent write window exists.
- `cdc.EventType` widening (`pr_review_thread_added`/`_resolved`) is safe — no exhaustive switch over `EventType` exists in the tree.

whoisasx:
@harshitsinghbhandari thanks for the high-recall pass. I pushed `a5f96a9` with the correctness fixes that were actionable in this PR, plus regression coverage. Per finding:

1. **`CI.HeadSHA` empty in SCM → legacy reaction projection** — fixed. `scmToPRObservation` now falls back to `PR.HeadSHA` when `CI.HeadSHA` is absent, so CI nudge dedup signatures remain commit-specific on review-only/local-reconstructed observations. Added `TestSCMObservationUsesPRHeadWhenCIHeadMissing`.

2. **`InsertLegacyPRComment` / legacy replacement semantics** — the existing path did delete unthreaded legacy comments before inserting, so edited legacy comments were not actually stranded by `INSERT OR IGNORE`. I added explicit regression coverage (`TestWritePRReplacesLegacyCommentBodies`) to lock that behavior down. I also fixed the related real collision risk: legacy `WritePR` now uses a narrow `UpsertLegacyPR` that updates only legacy scalar columns and preserves SCM-owned metadata/hashes/timestamps from `WriteSCMObservation`.

3. **Crash after lifecycle success but before hash acknowledgement** — acknowledged as a remaining at-least-once-delivery edge. The current two-phase write intentionally prevents the worse failure mode (lost lifecycle effects after a failed delivery/restart). Exactly-once behavior across a crash after successful lifecycle delivery would require a durable reaction/outbox acknowledgement model, which is larger than this PR and should be tracked separately.

4. **Review-thread second page trigger** — this was a naming/readability issue, not the “newest unresolved” behavior. GitHub returns `last:N` connection nodes in connection order, so `latest[0]` is the oldest thread in the latest window. I added an explicit comment and `oldestLatestUnresolved` variable to make the boundary condition clear. Existing provider tests cover both “oldest resolved → no fallback” and “oldest unresolved → one older page”.

5. **Repeated semantic hash computation** — valid efficiency note, but not correctness-blocking. I left this unchanged to avoid widening the patch; the current cost is bounded and the next clean refactor is to carry prepared hash values through `domainFromObservation`.

6. **`enrichFailureLogs` O(M×N) matching** — fixed. It now pre-indexes checks by `(name, providerID)` and updates matching check rows directly while preserving duplicate-key behavior.

7. **Legacy/SCM mixed write path cleanup** — partially addressed. I kept the legacy path because tests and the legacy service still compile against it, but separated the legacy PR upsert so it no longer touches SCM-owned observer fields. Full legacy deletion can be a separate cleanup PR.

8. **Duplicate FIFO cache helpers** — valid cleanup suggestion, but no correctness issue. I did not fold all helpers in this patch; I added only the needed generic `cacheDelete` helper for the retry-cache fix below.

9. **`ReviewRefreshFailed` false entries holding FIFO slots** — fixed. A successful review refresh now deletes the retry-cache entry and removes it from the order slice. Added `TestPoll_SuccessfulReviewRefreshClearsRetryCacheSlot`.

10. **Provider-specific raw fields in domain** — valid architecture note, but this PR intentionally persists raw provider diagnostics alongside normalized fields so the observer can preserve/debug GitHub mergeability facts. Moving that to an opaque provider payload is a broader schema/API cleanup and should be separate from this migration PR.

Local validation run after the patch:

- `go test ./...`
- `go vet ./...`
- `go test -race ./...`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 cache clean && go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2 run --path-mode=abs`
- `go test -tags e2e -v ./internal/cli/...`

harshitsinghbhandari:
Let's merge this PR @whoisasx

PR #105, opened 2026-06-04
feat(terminal): wire CDC session-state streaming through /mux WebSocket

### LINKED ISSUE #104: feat(terminal): wire CDC session-state streaming to /mux WebSocket (live sessionPatch delivery)
## Problem

The `/mux` WebSocket delivers terminal PTY data and a session-state channel (`ch: sessions`) to the frontend `MuxProvider`. The PTY path works. The session-state path did not: the server was sending thin CDC notification objects (`seq`, `sessionId`, `eventType`) that do not match the `SessionPatch` shape (`id`, `status`, `activity`, `attentionLevel`, `lastActivityAt`) the TypeScript consumer expects. As a result, the dashboard never received live session-state updates over the WebSocket -- it had to poll or wait for a page refresh.

## Root cause

`terminal.Manager` subscribed to the CDC broadcaster but forwarded raw `cdc.Event` fields directly onto the wire. There was no concept of a `SessionSource` interface to resolve derived session state (PR facts + canonical lifecycle -> display status), and no `attentionLevel` mapping on the server side.

## What this issue tracks

Wire the full CDC -> derived-state -> `SessionPatch` delivery path through `/mux`:

- Define a `SessionSource` interface in the terminal package (clean boundary: terminal does not import the service layer)
- Implement `DeriveStatus` in the `domain` package (it only uses domain types; pulling it out of the service layer removes the circular pressure)
- Implement `attentionLevel()` server-side, mirroring `getDetailedAttentionLevel` in `packages/web/src/lib/types.ts`
- Build `daemonSessionSource` in the daemon package: wraps the sqlite store + `GetDisplayPRFactsForSession` + `DeriveStatus`
- Move DB enrichment off the CDC broadcaster hot path (the broadcaster contract explicitly requires non-blocking callbacks)
- Close the subscribe/snapshot gap so no event is missed during initial connect
- Add full unit coverage for the attention-level mapping and functional integration tests for the end-to-end wire path

Pritom14:
Closes #104

## What changed and why

Before this PR, the `/mux` WebSocket session-state channel forwarded raw CDC notification fields (`seq`, `projectId`, `sessionId`, `eventType`) directly onto the wire. The TypeScript `MuxProvider` consumer expects a `SessionPatch` shape (`id`, `status`, `activity`, `attentionLevel`, `lastActivityAt`). The mismatch meant the dashboard never received live session-state updates -- it could only poll or wait for a page refresh.

This PR closes that gap end-to-end.

---

## Flow

```
DB INSERT/UPDATE
(sessions / pr / pr_checks)
        |
   [DB trigger -- atomic with the write]
        |
    change_log
        |
  cdc.Poller (100ms tick)
        |
  cdc.Broadcaster.Publish(Event)
        |
        +---> terminal.Manager subscriber callback
                  |
           [non-blocking: push SessionID onto per-conn channel]
                  |             (broadcaster contract: must not block)
                  |
         writer goroutine (writeLoop)
                  |
        sessionSrc.Session()  <--------+
                  |                    |
          domain.DeriveStatus()        sqlite store
                  |                    |
        attentionLevel()      GetDisplayPRFactsForSession()
                  |
     serverMsg{ ch:"sessions", type:"snapshot",
                sessions:[{ id, status, activity,
                            attentionLevel, lastActivityAt }] }
                  |
        WebSocket frame to client
                  |
           MuxProvider.tsx
```

On subscribe, a full snapshot of all current sessions is queued through the same ordered channel before any per-session patches, so the client always starts with complete state.

---

### `internal/terminal/patches.go` (new)

Defines `SessionSource` (the terminal package's view of session state -- no service-layer import), `attentionLevel()` (server-side port of `getDetailedAttentionLevel` from `packages/web/src/lib/types.ts`), and `toSessionPatch` / `toSessionPatches` helpers.

### `internal/terminal/protocol.go`

Replaced the thin `sessionUpdate` struct with `sessionPatch` carrying the full wire shape the TS consumer expects. Updated `serverMsg.Sessions` accordingly.

### `internal/terminal/manager.go`

Three changes, each fixing a separate problem:

**1. Off-hot-path enrichment.** Added `sessionEvents chan sessionEvent` to `connState`. The CDC subscriber callback now pushes only a `SessionID` (non-blocking); all DB reads happen in `writeLoop`, off the broadcaster's serialized `Publish` loop. This honors the documented contract: "fn is called synchronously from the poller loop, so it must not block."

**2. No-gap subscribe/snapshot ordering.** Subscribe is registered before the initial snapshot is queued. Both the snapshot (full read) and per-session patches (single-session read) flow through the same ordered `sessionEvents` channel and are resolved live at write time -- so the last frame the client sees for any session always reflects current state, regardless of interleaving.

**3. Connection context for DB reads.** The writer uses the connection-scoped `ctx`, so a disconnect cancels any in-flight query.

### `internal/domain/status.go`

Moved `DeriveStatus` and `prPipelineStatus` from `internal/service/session/status.go` into the domain package. They use only domain types; keeping them in the service layer created coupling pressure. `internal/service/session/status.go` now delegates to the domain function.

### `internal/daemon/session_source.go` (new)

`daemonSessionSource` implements `terminal.SessionSource` over the sqlite store. Per session: loads the record, fetches PR facts, calls `domain.DeriveStatus`. Wired into `terminal.NewManager` via `WithSessionSource` in `daemon.go`.

### `internal/httpd/terminal_mux.go`

Exported `TerminalMuxHandler` (was `terminalMuxHandler`) so the integration test can mount it on `httptest.Server` without importing the full httpd stack.

---

## Tests

**Unit (`internal/terminal/patches_test.go`)** -- table-driven coverage of all 16 `attentionLevel` branches (every status constant + activity fallbacks + priority-ordering cases), `toSessionPatch` field shape including UTC timestamp normalization, and `toSessionPatches` ordering.

**Unit (`internal/terminal/manager_test.go`, `protocol_test.go`)** -- updated wire-shape golden test, subscribe/snapshot/CDC-forward tests with in-memory fake conn.

**Functional (`internal/integration/mux_streaming_functional_test.go`)** -- real sqlite store, real CDC poller, real `terminal.Manager`, real WebSocket via `httptest.Server` + `coder/websocket.Dial`:

- `TestMuxStreamingSessionPatchDelivery`: subscribe (empty snapshot) -> spawn -> poll CDC -> exact `idle`/`idle`/`working` patch on the wire
- `TestMuxStreamingPatchReflectsPRDerivedStatus`: apply CI failure -> poll CDC -> `ci_failed`/`review` patch on the wire (per-event enrichment path with non-trivial status mapping)
- `TestMuxStreamingInitialSnapshotContainsExistingSessions`: sessions in the store before connect appear in the first snapshot

All tests pass under `-race`. No new failures introduced.

### DIFF
--- a/backend/internal/daemon/daemon.go
+++ b/backend/internal/daemon/daemon.go
@@ -73,7 +73,8 @@ func Run() error {
 		}
 	}
 	runtimeAdapter := zellij.New(zellij.Options{SocketDir: zellijSocketDir})
-	termMgr := terminal.NewManager(runtimeAdapter, cdcPipe.Broadcaster, log)
+	termMgr := terminal.NewManager(runtimeAdapter, cdcPipe.Broadcaster, log,
+		terminal.WithSessionSource(&daemonSessionSource{store: store}))
 	defer termMgr.Close()
 
 	// Bring up the Lifecycle Manager and the reaper first: it makes the session
--- a/backend/internal/daemon/session_source.go
+++ b/backend/internal/daemon/session_source.go
@@ -29,14 +29,6 @@ func (s *daemonSessionSource) AllSessions(ctx context.Context) ([]domain.Session
 	return out, nil
 }
 
-func (s *daemonSessionSource) Session(ctx context.Context, id domain.SessionID) (domain.Session, bool, error) {
-	rec, ok, err := s.store.GetSession(ctx, id)
-	if err != nil || !ok {
-		return domain.Session{}, ok, err
-	}
-	return s.toSession(ctx, rec), true, nil
-}
-
 func (s *daemonSessionSource) toSession(ctx context.Context, rec domain.SessionRecord) domain.Session {
 	pr, ok, _ := s.store.GetDisplayPRFactsForSession(ctx, rec.ID)
 	if ok {
--- a/backend/internal/daemon/session_source.go
+++ b/backend/internal/daemon/session_source.go
@@ -1,0 +1,46 @@
+package daemon
+
+import (
+	"context"
+
+	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
+	"github.com/aoagents/agent-orchestrator/backend/internal/storage/sqlite"
+	"github.com/aoagents/agent-orchestrator/backend/internal/terminal"
+)
+
+// daemonSessionSource implements terminal.SessionSource over the sqlite store.
+// It resolves each session's derived display Status (record + PR facts) so the
+// terminal manager can build SessionPatch frames without importing the service layer.
+type daemonSessionSource struct {
+	store *sqlite.Store
+}
+
+var _ terminal.SessionSource = (*daemonSessionSource)(nil)
+
+func (s *daemonSessionSource) AllSessions(ctx context.Context) ([]domain.Session, error) {
+	recs, err := s.store.ListAllSessions(ctx)
+	if err != nil {
+		return nil, err
+	}
+	out := make([]domain.Session, 0, len(recs))
+	for _, rec := range recs {
+		out = append(out, s.toSession(ctx, rec))
+	}
+	return out, nil
+}
+
+func (s *daemonSessionSource) Session(ctx context.Context, id domain.SessionID) (domain.Session, bool, error) {
+	rec, ok, err := s.store.GetSession(ctx, id)
+	if err != nil || !ok {
+		return domain.Session{}, ok, err
+	}
+	return s.toSession(ctx, rec), true, nil
+}
+
+func (s *daemonSessionSource) toSession(ctx context.Context, rec domain.SessionRecord) domain.Session {
+	pr, ok, _ := s.store.GetDisplayPRFactsForSession(ctx, rec.ID)
+	if ok {
+		return domain.Session{SessionRecord: rec, Status: domain.DeriveStatus(rec, &pr)}
+	}
+	return domain.Session{SessionRecord: rec, Status: domain.DeriveStatus(rec, nil)}
+}
--- a/backend/internal/domain/status.go
+++ b/backend/internal/domain/status.go
@@ -19,3 +19,48 @@ const (
 	StatusIdle             SessionStatus = "idle"
 	StatusTerminated       SessionStatus = "terminated"
 )
+
+// DeriveStatus computes the user-facing display status from a SessionRecord and
+// optional PR facts. pr nil means the session has no associated PR.
+func DeriveStatus(rec SessionRecord, pr *PRFacts) SessionStatus {
+	if rec.IsTerminated {
+		if pr != nil && pr.Merged {
+			return StatusMerged
+		}
+		return StatusTerminated
+	}
+	if rec.Activity.State == ActivityWaitingInput {
+		return StatusNeedsInput
+	}
+	if pr != nil {
+		if pr.Merged {
+			return StatusMerged
+		}
+		if !pr.Closed {
+			return prPipelineStatus(*pr)
+		}
+	}
+	if rec.Activity.State == ActivityActive {
+		return StatusWorking
+	}
+	return StatusIdle
+}
+
+func prPipelineStatus(pr PRFacts) SessionStatus {
+	switch {
+	case pr.CI == CIFailing:
+		return StatusCIFailed
+	case pr.Draft:
+		return StatusDraft
+	case pr.Review == ReviewChangesRequest || pr.ReviewComments:
+		return StatusChangesRequested
+	case pr.Mergeability == MergeMergeable:
+		return StatusMergeable
+	case pr.Review == ReviewApproved:
+		return StatusApproved
+	case pr.Review == ReviewRequired:
+		return StatusReviewPending
+	default:
+		return StatusPROpen
+	}
+}
--- a/backend/internal/httpd/terminal_mux.go
+++ b/backend/internal/httpd/terminal_mux.go
@@ -25,13 +25,13 @@ func mountTerminalMux(r chi.Router, mgr *terminal.Manager, log *slog.Logger) {
 	if mgr == nil {
 		return
 	}
-	r.Get("/mux", terminalMuxHandler(mgr, log))
+	r.Get("/mux", TerminalMuxHandler(mgr, log))
 }
 
-// terminalMuxHandler upgrades the request to a WebSocket and hands the connection to the
+// TerminalMuxHandler upgrades the request to a WebSocket and hands the connection to the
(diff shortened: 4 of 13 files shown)


Pritom14:
## Follow-up: subscribe handshake fix

While verifying the end-to-end path against the frontend, I found that the `/mux` subscribe handshake did not match what the real client sends, so the sessions channel never actually started for the dashboard. This commit fixes that.

### What was wrong

The frontend subscribes by sending only a channel and a topics list, with no message type:

```
{ "ch": "subscribe", "topics": ["sessions", "notifications"] }
```

On the server, `handleSubscribe` required `msg.Type == "subscribe"` and `clientMsg` had no `topics` field at all. So:

1. `msg.Type` was empty, the guard returned early, and `Subscribe` was never called.
2. No initial snapshot was queued and no CDC patches were forwarded.

The functional tests passed only because they sent a synthetic `{"ch":"subscribe","type":"subscribe"}` frame that the real client never sends.

### The fix

Gate the subscription on the topics array the frontend already sends, not on the message type.

```
client                          server (handleSubscribe)
------                          ------------------------
{ch:"subscribe",                if events != nil &&
 topics:["sessions",   ───────▶    slices.Contains(topics, "sessions"):
          "notifications"]}            Subscribe(...)        // CDC fan-out
                                       queue full snapshot   // immediate state
                                else:
                                       no-op                 // no session work
```

- Added a `Topics []string` field to `clientMsg`.
- Subscribe to the session feed only when `topics` contains `"sessions"`.
- Removed the now-unused `msgSubscribe` constant.
- Updated the functional tests to send the real frontend frame (`topics` array, no `type`).

### Why gate on topics rather than just sending a type

Honoring topics is strictly more capable than making the client send a `type`. The frontend already declares intent per channel, so gating on it lets a client that only wants notifications skip all session enrichment (the per-event DB read plus JSON write in the write loop). The alternative would have left `topics` as a dead field while the server fanned out session work to every connection.

### Tests

- `TestServeSubscribeWithoutSessionsTopicSendsNoSnapshot`: a client subscribing to only `notifications` gets no snapshot and no forwarded CDC events.
- `TestClientMsgSubscribeFrameDecodes`: the real frontend frame (`ch` + `topics`, no `type`) decodes correctly.
- Existing snapshot, CDC forwarding, and the three `TestMuxStreaming*` functional tests updated to the real frame and passing.

### Still open (separate decision needed)

Single-session CDC patches are currently sent as `type:"snapshot"` with a one-element array, and the frontend's `setSessions(msg.sessions)` replaces the whole array. After this handshake fix, the first single-session patch would drop every other session from the client's view. Fixing that requires either always sending full snapshots (server-only, less efficient) or adding a `type:"patch"` with a merge-by-id on the frontend. That touches the frontend, so it is out of scope for this commit and tracked separately.

Pritom14:
## Follow-up: full snapshots instead of single-session patches

This closes the second issue noted above (single-session updates clobbering the client's session list), without any frontend change.

### What was wrong

The frontend treats a sessions/snapshot frame as the complete session list:

```
// MuxProvider.tsx
if (msg.type === "snapshot") setSessions(msg.sessions);  // replaces the whole array
```

The writer, though, sent single-session CDC updates as a one-element snapshot. On the client that meant the first CDC event after subscribe replaced the entire list with just one session, dropping every other session from view.

### The fix

Always emit a full snapshot of all sessions on any session-state signal, rather than a single-session patch.

```
CDC event  ──▶ push signal ──▶ writer: drain+coalesce signals
                                       AllSessions()  (one read)
                                       send {type:"snapshot", sessions:[...all]}
```

This keeps the existing frontend correct with zero client-side change, which matches the protocol's stated goal of letting the existing client connect unchanged.

### Why this over a "patch" message type

A `type:"patch"` plus client-side merge-by-id would be more bandwidth-efficient, but it requires a coordinated change in the separate frontend repo. Full snapshots stay entirely in this repo and are correct against the current client. The read amplification this introduces is bounded by coalescing: a burst of CDC events drains into a single `AllSessions` read per write cycle (`drainSignals`), and the read is a local single-user SQLite query off the broadcaster hot path.

### Cleanup

Removed the now-dead single-session path: `SessionSource.Session` and its implementations on the daemon source and the test fakes. The interface is now just `AllSessions`.

### Tests

All terminal and integration tests pass (race-clean), including the three `TestMuxStreaming*` functional tests and the topic-gating tests from the previous commit. The only failure is the pre-existing `TestSessionStreamsRealZellijPane`, which is an environmental zellij socket-path limit unrelated to this change.

neversettle17-101 (commented):
Leaving the `/mux`-vs-dedicated-endpoint question aside (taking that separately) — two things about *this* change that I'd want addressed before merge. Both are about where the new DB work lands.

### 1. The session snapshot read runs on the terminal's single write goroutine

In `writeLoop` (`backend/internal/terminal/manager.go`), the new `sessionEvents` case calls `sessionSrc.AllSessions(ctx)` synchronously:

```go
case <-c.sessionEvents:
    drainSignals(c.sessionEvents)
    all, err := c.mgr.sessionSrc.AllSessions(ctx)   // synchronous DB read
    ...
    c.conn.WriteJSON(ctx, serverMsg{Ch: chSessions, Type: msgSnapshot, Sessions: toSessionPatches(all)})
```

`writeLoop` is the single writer for the whole connection, including PTY output drained from `c.out`. So this DB read now sits inline on the path that flushes terminal bytes — a slow or contended snapshot read stalls PTY output for that connection.

The PR description's goal was to move DB reads "off the broadcaster's hot path," and it does correctly get them off the *poller* loop. But it lands them on the *write* loop, which is the latency-sensitive path for the terminal. Suggest resolving the snapshot off `writeLoop` (e.g. a small dedicated goroutine that feeds resolved patches into `c.out`) so a slow read can't block PTY flushing.

### 2. The snapshot read is N+1 per session, and it fires per CDC event per connection

`AllSessions` (`session_source.go`) calls `GetDisplayPRFactsForSession` once per session, which is a single-session query (`backend/internal/storage/sqlite/store/pr_facts.go:16`). So each snapshot is O(sessions) queries. Combined with #1 and the full-snapshot-per-event model (every CDC event re-reads and re-ships *all* sessions to *every* connected client), this is the part that bites first as session count grows. The coalescing helps batch bursts but doesn't bound the per-snapshot cost. Worth either a batched PR-facts query, or at minimum a comment acknowledging the fan-out as a known scaling limit.

Everything else looks good — the `DeriveStatus` move into `domain` is a clean call, the subscribe-before-snapshot ordering is right, and the functional tests are thorough.

Pritom14:
closing this

PR #118, opened 2026-06-05
feat(agents): agent platform — registry, activity hooks, harness allowlist

yyovil:
## Agent platform (base of a stacked series)

This is the **root PR** of a stack. It lands the shared platform that every
per-agent adapter plugs into; each subsequent PR in the stack adds **one agent
adapter** on top of this.

Wired here for the three already-shipped harnesses — `claude-code`, `codex`,
`opencode`.

### What's in it

- **`adapters/agent/registry`** — single source of truth for the shipped
  adapters (`Constructors()`); the daemon resolves a session's harness through
  it. Adding an agent becomes one line here.
- **`adapters/agent/activitydispatch` + the `ao hooks` command** — maps an
  agent's native hook callbacks onto AO activity states
  (`active`/`idle`/`waiting_input`/…).
- **claude-code / codex / opencode** — emit `SessionStart` /
  `UserPromptSubmit` / `Stop` activity through the dispatcher.
- **HTTP + OpenAPI** — report session activity state.
- **db** — a *single* migration that widens `sessions.harness` to all shipped
  harnesses at once, replacing what was an 18-migration chain where each step
  textually depended on the previous one's output (and silently no-op'd if
  merged out of order). New adapters now need **no** further migration.
- **domain** — harness constants + an `--agent` alias for `ao spawn`.

### Why a stack

The per-agent adapters are independent packages, but they all register through
the same handful of files (`Constructors()`, the activity dispatcher, the
harness allowlist). Landing the platform once, then stacking one adapter per PR,
keeps each adapter reviewable in isolation without a tangle of migration-order
and merge hazards.

### Testing

`go build` / `go vet` / `go test ./...` green, except the pre-existing
`TestSessionStreamsRealZellijPane` integration test, which fails only on the
known environmental zellij IPC-socket-path-too-long issue (long macOS `$TMPDIR`)
— unrelated to this change.


yyovil:
Superseded by https://github.com/aoagents/ReverbCode/pull/119 — recreating in-upstream so the adapter stack can base off it.

PR #165, opened 2026-06-09
feat: add workspace project registration foundation

harshitsinghbhandari:
## Summary
- add workspace project persistence (`projects.kind`, `workspace_repos`, `session_worktrees`) plus store/domain models
- add `ao project add --as-workspace` and daemon registration for root-as-repo parent setup
- validate direct child repos, gitignore children before parent commits, and block parent gitlinks
- regenerate OpenAPI and frontend API types

Implements the unblocked registration/schema slices from discussion #164.

## Tests
- `cd backend && go build ./... && go test -race ./...`

Note: `npm run frontend:typecheck` is currently blocked by missing frontend deps/types (`electron`, `fumadocs-*`, Next RequestInit type) in this worktree.

## Review hardening (aec5bc7)
- rebased onto `main` (conflicts in service.go/types.go resolved; OpenAPI + frontend schema regenerated)
- parent now gets the same safety checks as children: linked-worktree parents (`WORKSPACE_PARENT_IS_WORKTREE`, with `--separate-git-dir` repos still accepted) and bare parents (`WORKSPACE_PARENT_BARE`) are rejected before any mutation
- failed init rolls back: the created `.git` is removed and the pre-call `.gitignore` restored; `guardNoGitlinks` errors carry an actionable `git rm --cached` suggestedFix
- child repos named `__root__` rejected (`WORKSPACE_CHILD_RESERVED_NAME`); dead `__root__` ordering removed from `ListWorkspaceRepos`
- `Service.addMu` serializes `Add`, closing the registration TOCTOU race
- 8 new tests: adopt-existing-parent, worktree parent/child rejection, bare parent, reserved name, init rollback + deterministic retry, concurrent same-path race, separate-git-dir parent


backend/internal/service/project/workspace_registration.go
```diff
@@ -0,0 +1,241 @@
...
+	children, err := detectWorkspaceChildren(ctx, parent, projectID, registeredAt)
+	if err != nil {
+		return nil, err
+	}
+	if len(children) == 0 {
+		return nil, apierr.Invalid("WORKSPACE_REPOS_REQUIRED", "Workspace project must contain at least one direct child git repository", map[string]any{
+			"suggestedFix": "Create or move child repositories directly under the workspace folder, then retry.",
+		})
+	}
+	if isGitRepo(parent) {
```
harshitsinghbhandari, line 44: **Parent lacks the safety checks children get.** `validateWorkspaceChild` rejects worktree children (`.git` file) and bare repos, but the parent only passes through `isGitRepo` — which compares `rev-parse --show-toplevel` to the path, and a **linked worktree's** toplevel *is* the worktree root. So a parent that is a worktree of an external repo takes the adopt path, and `adoptWorkspaceParent` commits `.gitignore` changes directly onto that external repo's checked-out branch.

Mirror the child checks here: `Lstat(parent/.git).IsDir()` and the bare-repo probe, with a `WORKSPACE_PARENT_IS_WORKTREE` error. It's ~10 lines and closes the one path where AO silently writes commits into a repo the user didn't point it at.

backend/internal/service/project/workspace_registration.go
```diff
@@ -0,0 +1,241 @@
...
+	if _, err := gitOutput(ctx, parent, "init", "-b", domain.DefaultBranchName); err != nil {
+		return apierr.Invalid("WORKSPACE_PARENT_INIT_FAILED", "Failed to initialize workspace parent git repository", map[string]any{"error": err.Error()})
+	}
+	if _, err := ensureWorkspaceGitignore(parent, repos); err != nil {
+		return apierr.Invalid("WORKSPACE_PARENT_GITIGNORE_FAILED", "Failed to write workspace parent .gitignore", map[string]any{"error": err.Error()})
+	}
+	if _, err := gitOutput(ctx, parent, "add", "-A"); err != nil {
+		return apierr.Invalid("WORKSPACE_PARENT_ADD_FAILED", "Failed to stage workspace parent files", map[string]any{"error": err.Error()})
+	}
+	if err := guardNoGitlinks(ctx, parent); err != nil {
```
harshitsinghbhandari, line 270: **Failure here strands the user with no recovery path.** Sequence: parent is a plain folder containing a *nested* repo at depth ≥2 (e.g. `packages/foo/.git` — `detectWorkspaceChildren` only scans depth 1, so it's neither registered nor gitignored). Then:

1. `git init` + `.gitignore` write + `git add -A` stages `packages/foo` as a gitlink → `guardNoGitlinks` errors. Registration fails, but the folder is now a half-initialized repo with a staged gitlink and no commit.
2. Every retry takes the **adopt** path, where `ensureWorkspaceGitignore` reports unchanged → early return → the outer `guardNoGitlinks` in `prepareWorkspaceProject` hits the same stale staged entry. Permanent failure until the user manually runs `git rm --cached`.

Options, roughly in order of preference: (a) on init-path failure, roll back (`rm -rf .git` of the repo this code just created — safe because this branch only runs when the parent was *not* a repo); (b) instead of erroring, `git rm --cached` the gitlink paths and add them to `.gitignore`; (c) at minimum, put the exact remediation command in `suggestedFix` — the current message says "child repos must be gitignored" which doesn't apply to nested repos AO will never ignore.

backend/internal/service/project/workspace_registration.go
```diff
@@ -0,0 +1,241 @@
...
+	entries, err := os.ReadDir(parent)
+	if err != nil {
+		return nil, apierr.Invalid("INVALID_PATH", "Workspace path could not be read", nil)
+	}
+	var repos []domain.WorkspaceRepoRecord
+	for _, entry := range entries {
+		if !entry.IsDir() {
+			continue
+		}
+		name := entry.Name()
```
harshitsinghbhandari, line 151: **Reserved name not enforced.** A child directory literally named `__root__` is accepted here and registered into `workspace_repos`. Once the spawn slice lands, `session_worktrees` keys the root worktree on `RootWorkspaceRepoName = "__root__"` with `PRIMARY KEY (session_id, repo_name)` — a real child by that name silently collides with the root row.

One line now avoids a data-corruption class later:
```go
if name == string(domain.RootWorkspaceRepoName) { return nil, apierr.Invalid("WORKSPACE_CHILD_RESERVED_NAME", ...) }
```

harshitsinghbhandari (commented):
## Review — workspace project registration foundation

Reviewed the full diff (service, registration flow, sqlite migration/queries/store, OpenAPI/CLI/frontend types). CI is green. The overall structure is solid: transactional child-set replacement in `UpsertWorkspaceProject`, gitlink guard runs **before** commit in both paths, pathspec-limited commit in adopt mode (verified this works on unborn HEAD too), DTO drift e2e covers `asWorkspace`, migration follows existing multi-statement conventions, deterministic sorted child order.

Since this feature mutates user repositories, I focused on failure/edge paths. Three substantive findings (inline), plus smaller notes below.

### Substantive (inline comments)
1. **Parent gets none of the safety checks children get** — a linked worktree parent passes `isGitRepo` and the adopt path will commit onto an external repo's branch.
2. **Failed init leaves the parent half-mutated with no recovery** — a nested repo at depth ≥2 trips `guardNoGitlinks` after `git init` + `add -A`, and every retry then fails forever on the stale staged gitlink.
3. **Reserved `__root__` child name not rejected** — collides with `RootWorkspaceRepoName` in `session_worktrees` (PK `session_id, repo_name`) once spawning lands.

### Medium
- **Concurrent `Add(AsWorkspace)` on the same path races**: both calls pass `FindProjectByPath`, then race in `git init`/commit. The store has `writeMu` but the new service-level filesystem mutation is unguarded — the existing Add never touched disk, so this is new exposure. A per-path (or global) mutex around `prepareWorkspaceProject` would close it.
- **Adopt path is untested**: tests cover init-from-plain-folder and unborn-child rejection, but the riskier path — committing into a user's *existing* parent repo — has no coverage. Neither do the gitlink guard or the worktree-child rejection. Given how delicate this is, those three tests are worth adding before the spawn slice builds on this.

### Minor / nits
- `ListWorkspaceRepos` SQL orders `__root__` first, but the `WorkspaceRepoRecord` doc comment says the table holds only direct children (root is never stored). Dead clause or stale comment — pick one. (In `ListSessionWorktrees` the same clause *is* meaningful.)
- `ensureWorkspaceGitignore` writes `.gitignore` with `0o600` — convention is `0o644`; a private-permission `.gitignore` in a shared checkout is odd.
- In adopt mode the denylist (`node_modules/`, `dist/`, …) is appended and committed into the user's **existing** repo — beyond what's needed to ignore children. Consider adding only the child ignores when adopting.
- `resolveGitOriginURL(path)` is called twice on the workspace path (row construction + after prepare); the first is wasted.
- OpenAPI `kind` is a bare `string`; an enum (`single_repo`, `workspace`) would give clients real typing.
- `Degraded.Kind` is added and marked required in the spec, but nothing constructs `Degraded` today — when that path lands, remember to populate it or the wire value will be `""`.
- Gitignore dedup is exact-line: an existing `api/` entry won't dedupe against `/api/` (harmless duplicate only).

None of the nits are blocking. Findings 1–3 are edge cases but real, and this is the foundation other slices build on — I'd fix at least 1 and 3 (both are small) and decide explicitly on 2 before merge.

backend/internal/service/project/service.go
```diff
@@ -131,14 +135,33 @@ func (m *Service) Add(ctx context.Context, in AddInput) (Project, error) {
...
 		ID:            string(id),
 		Path:          path,
 		RepoOriginURL: resolveGitOriginURL(path),
 		DisplayName:   name,
-		RegisteredAt:  time.Now(),
+		RegisteredAt:  registeredAt,
+		Kind:          domain.ProjectKindSingleRepo,
 		Config:        config,
 	}
+	if in.AsWorkspace {
```
harshitsinghbhandari, line 162: **TOCTOU race on concurrent registration.** Two concurrent `Add(AsWorkspace)` calls for the same path both pass the `FindProjectByPath` check above, then both run `prepareWorkspaceProject` — racing `git init`/`add`/`commit` against each other in the same directory. The store's `writeMu` serializes the DB writes but not these filesystem mutations (the pre-existing single-repo Add never mutated disk, so this exposure is new in this PR). A mutex held across prepare+upsert — or a per-path lock — closes it.

backend/internal/service/project/workspace_registration.go
harshitsinghbhandari: Fixed in aec5bc7 — added `validateWorkspaceParent`, called before any mutation: rejects linked-worktree parents (`.git` file with `--git-dir` ≠ `--git-common-dir`, so `--separate-git-dir` standalone repos still pass) as `WORKSPACE_PARENT_IS_WORKTREE`, and bare parents as `WORKSPACE_PARENT_BARE`. Covered by `TestManager_AddWorkspaceRejectsWorktreeParent`, `TestManager_AddWorkspaceRejectsBareParent`, and `TestManager_AddWorkspaceAdoptsSeparateGitDirParent`.

backend/internal/service/project/workspace_registration.go
harshitsinghbhandari: Fixed in aec5bc7 with option (a) + (c): `initWorkspaceParent` snapshots the pre-call `.gitignore` and, on any failure after a successful `git init`, removes the `.git` it created and restores `.gitignore` exactly (defer on named return; not registered until init succeeds, so a pre-existing repo is never deleted). `guardNoGitlinks` details now include a concrete `git rm --cached <path>` suggestedFix. `TestManager_AddWorkspaceInitRollsBackOnNestedGitlink` asserts the rollback and that a retry fails deterministically with the same error.

backend/internal/service/project/workspace_registration.go
harshitsinghbhandari: Fixed in aec5bc7 — `detectWorkspaceChildren` rejects a child repo named `domain.RootWorkspaceRepoName` with `WORKSPACE_CHILD_RESERVED_NAME` (placed after the `isGitRepo` check, so a plain folder named `__root__` is still just skipped). Test: `TestManager_AddWorkspaceRejectsReservedChildName`. Also removed the now-dead `__root__` ORDER BY clause from `ListWorkspaceRepos` (sqlc regenerated); `ListSessionWorktrees` keeps its meaningful one.

backend/internal/service/project/service.go
harshitsinghbhandari: Fixed in aec5bc7 — added `Service.addMu` held across the whole `Add` body (from before `FindProjectByPath` through the upsert), closing the TOCTOU for both the path and id conflict checks; comment in code explains the store's writeMu doesn't cover the new filesystem mutations. `TestManager_AddWorkspaceConcurrentSamePath` races two goroutines under `-race`: exactly one succeeds, the other gets `PATH_ALREADY_REGISTERED`.

harshitsinghbhandari:
Correction to one review nit: the `.gitignore` 0o600 → 0o644 suggestion was wrong — this repo's golangci-lint gosec (G306) policy enforces WriteFile permissions ≤0600, which is why the original code used 0o600. Kept at 0o600. All other findings addressed in 8ef4280; CI is fully green (9/9) after the rebase.

PR #166, opened 2026-06-10
chore: add prettier config and CI auto-formatter

yyovil:
## What

Adds Prettier config with CI-driven auto-formatting. Splits the formatter setup out of #156 which bundled it with the React renderer (see review comment there).

## How it works

- `.prettierrc` — the only hand-written config (120-char print width, double quotes, trailing commas)
- `.prettierignore` — excludes generated files (`schema.ts`, `openapi.yaml`), lockfiles, and `backend/` (Go uses `gofmt`)
- `.github/workflows/prettier.yml` — on every push to a non-main branch, runs `prettier --write .` and commits any changed files back via `GITHUB_TOKEN`

**No local enforcement.** No husky, no lint-staged. Developers never need to run Prettier. The formatted output shows up as a CI commit on the branch, not as local uncommitted changes.

GitHub Actions does not re-trigger workflows on commits made with `GITHUB_TOKEN`, so there is no feedback loop.

## After merge

The first push to any feature branch will trigger a formatting pass over the whole codebase. That single CI commit is the mass-reformat — it comes from the bot, not from any developer's machine.


.envrc
```diff
@@ -1 +1,3 @@
 use flake
+entire enable --agent claude-code
```
yyovil: Split this into a separate envrc file that stays only on my system and drop these suggested code changes. Rest of the team doesn't use entire cli for agent session audits.

.prettierrc
```diff
@@ -0,0 +1,9 @@
+{
+  "useTabs": false,
```
yyovil: plz useTabs. turn this to true.

.github/workflows/sync-fork.yml
yyovil, line 1: this doesn't belong here. It belongs to my fork only. out of PR's scope. remove this.

yyovil (commented):
@ashish921998 plz approve these changes.

PR #168, opened 2026-06-10
feat(frontend): scaffold for frontend with complimentary backend changes

yyovil:
## What

Rebuilds the Electron desktop frontend as a full **React 19 + TypeScript renderer** (Vite + electron-forge), replacing the skeleton frontend. This is a clean extraction of @ashish921998's work from #156 — rebased onto current `main`, without the Prettier mass-reformat / husky tooling that made that PR hard to review.

## Stack

| Layer | Choice |
|---|---|
| Shell | Electron + child_process daemon supervision (electron-forge) |
| UI | React 19 + TypeScript renderer (Vite, contextBridge preload) |
| Routing | TanStack Router (file-system routes; hash history for the `app://` renderer) |
| Data | TanStack Query + EventTransport — CDC SSE on `/api/v1/events` |
| API types | openapi-typescript + openapi-fetch off `openapi.yaml` |
| Styling | Tailwind + shadcn/ui |
| Layout | react-resizable-panels, sizes persisted to Zustand |
| Terminal | @xterm/xterm per-session PTY over `/mux` WebSocket + WebGL addon |
| Packaging | electron-forge + update-electron-app auto-updater |
| Testing | Vitest + RTL · Playwright |

## Backend additions

- **`cors.go`** — allowlist-only CORS middleware, never `*` or `null`; handles Private Network Access preflight for the `app://` renderer → loopback daemon
- **`session.TerminalHandleID`** exposed in domain + OpenAPI spec (+ controller tests)
- **`project.Path`** added to OpenAPI spec, service, store, and tests

## Design

`DESIGN.md` is the source of truth for the emdash-matched dark UI (tokens, blue accent, status glyph spec, orchestrator-led layout).

## Notes

- Single squashed commit on top of current `main`; landing site untouched; backend builds + `go vet` clean.
- All renderer implementation by @ashish921998 — this is a clean rebase of his work.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #170, opened 2026-06-11
fix(codex): reliable activity signals — session-flag hooks, trust bypass, no_signal watchdog

ashish921998:
## Why

Codex sessions spawned by AO showed a confident **"idle" forever while the agent was visibly working**, and spawns into never-trusted repos hung invisibly. Root causes, verified against codex-cli 0.136.0 and its source:

1. Codex never loads hook config from AO's per-session worktrees: project-local `.codex/` layers only load from **trusted** directories, and for linked git worktrees codex sources hook declarations from the matching folder in the **root checkout** — so the workspace `.codex/hooks.json` AO wrote was dead config.
2. `--dangerously-bypass-hook-trust` does not bypass the interactive "Do you trust this directory?" prompt, which blocked headless panes.
3. Near the account rate limit, codex's "switch to a cheaper model?" dialog blocks the first turn and swallows the auto-submitted spawn prompt.

This re-lands work stranded on the closed redesign PR #156 (cherry-picked codex fix) plus a watchdog so the *next* silent breakage of this class is visible.

## What

**Commit 1 — codex hook delivery + trust (verified by real spawn: idle → active → idle hands-free):**
- Hooks ride the launch/resume argv as `-c 'hooks.<Event>=[...]'` session-flag config (not trust-gated; aggregates with user hooks) alongside the existing `--dangerously-bypass-hook-trust`.
- `-c 'projects={"<worktree>"={trust_level="trusted"}}'` (inline-table form — the dotted `projects."<path>".trust_level` key is corrupted by codex's `-c` dot-splitting) suppresses the directory-trust prompt, scoped to the single invocation.
- `-c notice.hide_rate_limit_model_nudge=true` keeps the rate-limit dialog from hanging headless panes.
- `GetAgentHooks` writes nothing into worktrees anymore (they stay git-clean); it only strips legacy AO entries.

**Commit 2 — watchdog + hardening:**
- New `sessions.first_signal_at` (migration 0010): raw receipt of the first hook callback per spawn/restore. A live session with no receipt past a 90s grace derives the new **`no_signal`** display status instead of `idle`; frontend maps it to needs-you. CDC trigger updated so the transition pushes live.
- Session PATH pinned to the daemon binary so the bare `ao` in hook commands can't resolve to a foreign CLI; `ao hooks` failures land in `hooks.log`.
- `ao doctor` gains: `hooks-log` (warns on deliveries failed in the last 24h), `ao-binary` (PATH identity), and `codex-launch-flags` — a canary that probes the installed codex with the exact flags AO injects (exported by the adapter from the same builders as the spawn argv), so a codex CLI change that breaks injection becomes a doctor warning.
- Codex hook timeout 30s → 5s; agent PRD docs updated to the implemented callback flow.

## Verification

- `go build ./...`, `go test ./...` (50 packages) green; changed packages lint clean (golangci-lint v2.12.2).
- Live: real AO spawn of a codex worker into a fresh, never-trusted repo — activity transitioned idle → active → idle with zero human interaction; worktree contained no `.codex/` and stayed git-clean; `ao doctor` shows the new checks passing against codex-cli 0.136.0.
- Frontend: `workspace.test.ts` passes 26/26 (the only frontend file touched). Note: main's renderer suite has 7 pre-existing failing test files plus a vite-config typecheck error from the #168 scaffold, unaffected by this PR.

## Notes

- #169 (dirty-worktree teardown) touches `codex/hooks.go` too; whichever lands second resolves by keeping this PR's no-write codex adapter and dropping its `EnsureWorkspaceGitignore` call there (the conformance test passes vacuously for adapters that write nothing).
- Hook-derived metadata persistence (`agentSessionId` for `codex resume`) remains unimplemented and is now documented as such — follow-up candidate.

### DIFF
--- a/backend/internal/adapters/agent/activitydispatch/dispatch.go
+++ b/backend/internal/adapters/agent/activitydispatch/dispatch.go
@@ -57,3 +57,14 @@ func Derive(agent, event string, payload []byte) (domain.ActivityState, bool) {
 	}
 	return derive(event, payload)
 }
+
+// SupportsHarness reports whether a harness has an activity pipeline at all:
+// a registered deriver here means its adapter installs `ao hooks <harness>`
+// callbacks that can reach the daemon. Status derivation uses this to decide
+// whether prolonged silence is suspicious (no_signal) or simply all a hook-less
+// harness can ever report (idle). Harness names and `ao hooks` agent tokens are
+// the same strings by convention.
+func SupportsHarness(h domain.AgentHarness) bool {
+	_, ok := Derivers[string(h)]
+	return ok
+}
--- a/backend/internal/adapters/agent/activitydispatch/dispatch_test.go
+++ b/backend/internal/adapters/agent/activitydispatch/dispatch_test.go
@@ -1,0 +1,33 @@
+package activitydispatch
+
+import (
+	"testing"
+
+	"github.com/aoagents/agent-orchestrator/backend/internal/domain"
+)
+
+// Every deriver key must be a known harness name: SupportsHarness equates the
+// two, so a token that drifts from its harness constant would silently report
+// the harness as hook-less.
+func TestDeriverTokensAreKnownHarnesses(t *testing.T) {
+	for token := range Derivers {
+		if !domain.AgentHarness(token).IsKnown() {
+			t.Errorf("deriver token %q is not a known AgentHarness", token)
+		}
+	}
+}
+
+func TestSupportsHarness(t *testing.T) {
+	for _, h := range []domain.AgentHarness{domain.HarnessCodex, domain.HarnessClaudeCode, domain.HarnessOpenCode} {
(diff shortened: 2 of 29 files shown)


ashish921998:
Good catch on `TestSessionWorktreesRoundTrip` — its removal was **not intentional**. The branch ported `store_test.go` wholesale from the closed redesign branch, whose copy predates #165, silently dropping the test #165 added there. Restored in bd7009b: the file is now main's version plus only this branch's `TestSessionFirstSignalRoundTrip` addition; both round-trip tests pass.

Also audited every other ported file against main for the same failure mode — the remaining deletions are this branch's intended refactors (PATH-pin constructor changes, codex adapter rewrite, doc updates); no other main-side content was lost. #168's `TerminalHandleID` hunks in `domain/session.go` / `service/session/service.go` are present.

PR #180, opened 2026-06-11
feat(sessions): archive worker sessions from the sidebar row

ashish921998:
## What

Terminated workers can now be **archived** to declutter the sidebar without destroying the row, and **unarchived** to bring them back.

- **Sidebar row UX:** right-click a terminated worker → **Archive worker**. The row leaves the default list and lands behind a collapsed **"Archived (n)"** disclosure at the end of that project's session list (dimmed, still navigable). Right-click an archived row → **Unarchive worker**. Running workers get no menu — archive is terminated-only.
- **Daemon:** sessions gain a nullable `archived_at` (migration 0011, mirroring the projects soft-delete pattern); `POST /sessions/{id}/archive` and `/unarchive` routes; an `archived` filter on `GET /sessions`; `isArchived` on the Session read model.
- **Invariants:** archiving a running session 409s (`SESSION_NOT_TERMINATED`) and restore clears the flag, so an active agent can never be hidden. The full `UpdateSession` write never touches `archived_at`, so lifecycle writes can't clobber user intent. The sessions CDC update trigger is recreated so archive flips fan out `session_updated` events.
- **Renderer:** `useWorkspaceQuery` splits `archivedSessions` out of `sessions`, so the kanban board, sidebar counts, and every other default surface ignore archived workers automatically; `SessionView` still resolves them so archived rows open normally.

## Notes

- Stacked on #178 (base: `feat/wire-core-workflows`); retarget to `main` once that merges. The UI was originally built on the old emdash shell and re-landed on the new agent-orchestrator-clone shell in the merge commit.
- The `ui/context-menu` component (from #178) had `aria-hidden` on its overlay, hiding every menu item from the accessibility tree — fixed here; the new tests query menu items by role.
- Kill/restore row actions (lost with the old shell) are **not** re-added here — one issue per PR; they belong to the #178 follow-up that re-ports the lifecycle surfaces. The context-menu infra this PR adds is where they'll slot in.
- The CLI (`ao session list`) shows archived sessions with no marker; a CLI flag/column is a possible follow-up.

## Verified

- `npm run lint` (backend tests + golangci) green, `go test -race` on touched packages green, spec drift/route parity green, `npm run api` produces no diff
- Frontend: 123 vitest tests green (new: store/service/controller archive tests, query-split test, 3 Sidebar context-menu/disclosure tests), typecheck clean


harshitsinghbhandari:
I don't think we want to bring back archiving worker sessions. We can kill a session and restore it anyways so there is no need of archiving.

PR #209, opened 2026-06-12
fix(project): detect the repo's default branch on add

### LINKED ISSUE #208: ao project add records `main` as default branch for non-main repos, breaking ao spawn
## Bug

`ao project add` never inspects the repo's checked-out branch. The project's `Config.DefaultBranch` is left empty and `WithDefaults()` forces it to `"main"` (`backend/internal/domain/projectconfig.go:48`). For any repo whose default branch isn't literally `main` (e.g. `master`, `develop`, `trunk`), every subsequent `ao spawn` fails because the worktree base ref doesn't exist.

## Reproduction

1. `git init` a repo (no `-b`, so on a host without `init.defaultBranch` it lands on `master`), add one commit.
2. `ao project add --path <repo> --id x` → `ao project get x` reports `default branch: main`. ❌ (the repo is on `master`)
3. `ao spawn --project x --prompt hi` →
   ```
   BRANCH_NOT_FETCHED: "ao/x-1" has no local head, no remote, and no tag — run `git fetch` then retry
   ```
4. The identical repo on `main` spawns fine. `ao project add` exposes **no `--branch` flag**, so there is no CLI workaround.

## Root Cause

- `backend/internal/service/project/service.go` `Add` (single-repo path) registers the project without detecting the repo's branch, so `Config.DefaultBranch` stays empty → defaults to `main`.
- At spawn, `resolveBaseRef` (`backend/internal/adapters/workspace/gitworktree/workspace.go:253`) tries `origin/main`, `refs/heads/main`, and tags; none exist on a `master` repo → `errNoBaseRef` → `ErrBranchNotFetched`.
- The surfaced message ("run `git fetch`") is misleading: there is no remote, and the base branch name is simply wrong.

## Fix

Detect the repo's current branch with `git symbolic-ref --short HEAD` at registration and store it as `DefaultBranch` when the user didn't configure one and it diverges from `main`. Best-effort: a detached HEAD or git error falls back to the existing `main` default, so `add` never fails on this account.

## Impact

Breaks the README quickstart (`project add` → `spawn`) for every `master` / non-`main` repo — a large fraction of older and real-world repositories. No workaround via the CLI.

codebanditssss:
Closes #208.

## Problem

`ao project add` never inspected the repo's branch, so `Config.DefaultBranch` stayed empty and `WithDefaults()` forced it to `main`. Any repo whose default branch isn't `main` (`master`, `develop`, `trunk`, …) then failed **every** `ao spawn` with a misleading `BRANCH_NOT_FETCHED` — and `project add` has no `--branch` flag, so there was no CLI workaround. This breaks the README quickstart (`project add` → `spawn`) for a large class of repos.

## Change

- Add `resolveDefaultBranch` (mirrors `resolveGitOriginURL`: best-effort `git symbolic-ref --short HEAD`, never fails `add`).
- In `Service.Add` (single-repo path), record the detected branch as `DefaultBranch` **only** when the user set none and it diverges from `main` — so the common `main` repo keeps an empty (NULL) config and behavior is unchanged there. A detached HEAD or git error falls back to the existing `main` default.

## Test

- `TestManager_AddDetectsNonMainDefaultBranch`: a `master` repo records `master`; an explicit config still wins.
- Pinned the `gitRepo` test helpers to `git init -b main` so detection is deterministic regardless of the host's `init.defaultBranch` (this is also why two existing fixtures changed).
- `go build ./...`, `go vet ./...`, and the full `go test ./...` suite pass.
- Verified end-to-end against a live daemon: a fresh `master` repo now reports `default branch: master` on `project add` and `ao spawn` succeeds (`idle` session in zellij) where it previously failed `BRANCH_NOT_FETCHED`.

## Scope note

Surgical to the single-repo path. Workspace-root projects already `git init -b main`, so they are unaffected. The misleading "run `git fetch`" wording in the `BRANCH_NOT_FETCHED` message is left as-is — with correct detection it no longer fires for this case.

### DIFF
--- a/backend/internal/httpd/controllers/projects_test.go
+++ b/backend/internal/httpd/controllers/projects_test.go
@@ -509,7 +509,7 @@ func gitRepo(t *testing.T, name string) string {
 
 	}
 
-	if out, err := exec.Command("git", "init", dir).CombinedOutput(); err != nil {
+	if out, err := exec.Command("git", "init", "-b", "main", dir).CombinedOutput(); err != nil {
 
 		t.Fatalf("git init fixture: %v\n%s", err, out)
 
--- a/backend/internal/service/project/service.go
+++ b/backend/internal/service/project/service.go
@@ -194,6 +194,16 @@ func (m *Service) Add(ctx context.Context, in AddInput) (Project, error) {
 	if !isGitRepo(path) {
 		return Project{}, apierr.Invalid("NOT_A_GIT_REPO", "Repository path must point to a git repository", nil)
 	}
+	// Record the repo's actual checked-out branch as the project default so
+	// session worktrees base off a branch that exists. Without this a repo on
+	// `master` (or any non-`main` default) falls back to DefaultBranchName and
+	// every spawn fails BRANCH_NOT_FETCHED. Only persist when it diverges from
+	// the default, so the common `main` repo keeps an empty (NULL) config.
+	if row.Config.DefaultBranch == "" {
+		if branch := resolveDefaultBranch(path); branch != "" && branch != domain.DefaultBranchName {
+			row.Config.DefaultBranch = branch
+		}
+	}
 	row.RepoOriginURL = resolveGitOriginURL(path)
 	if err := m.store.UpsertProject(ctx, row); err != nil {
 		return Project{}, apierr.Internal("PROJECT_ADD_FAILED", "Failed to register project")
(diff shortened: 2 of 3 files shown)


harshitsinghbhandari:
Reviewed locally (build + `project` test suite pass). Approve. Surgical fix with the right root cause; nicely preserves the NULL-config path for `main` repos via the `!= DefaultBranchName` guard. One thing to consider: `resolveDefaultBranch` reads the checked-out branch (HEAD) rather than strictly the repo default, so adding a repo while on a feature branch would record that branch. Best-effort fallback is correct either way.

harshitsinghbhandari (approved):
Approve. Reviewed locally; build and the project test suite pass. Surgical fix with the right root cause.

PR #215, opened 2026-06-12
fix(session): stamp spawn timestamps in UTC

### LINKED ISSUE #214: ao session get shows created (local) and updated (UTC) in different timezones
## Bug

`ao session get` renders `created` and `updated` in **different timezones** for the same session: `created` in local time, `updated` in UTC.

## Reproduction

```
ao spawn --project <p> --prompt hi      # created+updated both local at first
ao session rename <id> "x"
ao session get <id>
# created: 2026-06-13T01:17:58+05:30      <- local (+05:30)
# updated: 2026-06-12T19:51:27Z           <- UTC (Z)
```

## Root Cause

The session manager's clock defaults to plain `time.Now` (local) in `New` (`backend/internal/session_manager/manager.go`), and `seedRecord` stamps `CreatedAt`/`UpdatedAt` from it at spawn. Every **other** session write uses `time.Now().UTC()` — rename (`service/session/service.go`), activity (`storage/sqlite/store/session_store.go`). So once any UTC-writing path touches the row, `created` (local) and `updated` (UTC) disagree. The CLI formats both with `time.RFC3339`, which faithfully preserves each value's zone.

## Fix

Default the session manager clock to UTC, matching the rest of the codebase. `created` and `updated` then render consistently.

## Impact

Confusing, inconsistent timestamps in `ao session get` and the JSON API; any consumer diffing the two fields sees a spurious offset. Labels: `bug`, `priority: low`.

codebanditssss:
Closes #214.

## Problem

`ao session get` showed `created` and `updated` in different timezones. **Two** clocks defaulted to local `time.Now` while the rest of the codebase writes UTC:

- the **session manager** clock → spawn-stamped `CreatedAt`/`UpdatedAt`
- the **lifecycle manager** clock → activity-driven `LastActivityAt`/`UpdatedAt`

A real spawn made the second one visible: after the first fix `createdAt` came back UTC, but `updatedAt`/`lastActivityAt` were still local (`+05:30`) the moment the agent reported activity.

## Change

Default both clocks to UTC, matching the codebase convention.

## Test

- `TestSpawn_StampsUTCTimestamps` (session manager): spawn-stamped `CreatedAt`/`UpdatedAt` are UTC.
- `TestMarkSpawned_StampsUTCActivity` (lifecycle): activity-driven `LastActivityAt` is UTC.
- `go build ./...`, full `go test ./...`, `gofmt`, and `golangci-lint` all clean.
- End-to-end: after spawn + activity, `ao session get` shows `created` and `updated` both ending in `Z`.

### DIFF
--- a/backend/internal/lifecycle/manager.go
+++ b/backend/internal/lifecycle/manager.go
@@ -52,7 +52,12 @@ type Manager struct {
 
 // New builds a Lifecycle Manager over the session store it writes and the messenger it uses for agent nudges.
 func New(store sessionStore, messenger ports.AgentMessenger, opts ...Option) *Manager {
-	m := &Manager{store: store, messenger: messenger, window: defaultRecentActivityWindow, clock: time.Now, react: newReactionState()}
+	// UTC so activity-driven LastActivityAt/UpdatedAt match spawn-stamped
+	// timestamps (the session manager clock is UTC too); a local clock here left
+	// `ao session get` showing created in UTC but updated in local time. A
+	// WithClock option may still override this in tests.
+	clock := func() time.Time { return time.Now().UTC() }
+	m := &Manager{store: store, messenger: messenger, window: defaultRecentActivityWindow, clock: clock, react: newReactionState()}
 	for _, opt := range opts {
 		opt(m)
 	}
--- a/backend/internal/lifecycle/manager_test.go
+++ b/backend/internal/lifecycle/manager_test.go
@@ -146,6 +146,21 @@ func TestMarkSpawnedStoresRuntimeMetadata(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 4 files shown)


harshitsinghbhandari:
Reviewed locally — approve. Right root cause: both clocks that stamp the displayed session timestamps now write UTC, and the tests genuinely catch the regression (`time.Local` is a distinct `*Location` from `time.UTC`). Build + touched-package tests pass. Two non-blocking notes: (1) the same local-default clock pattern survives in a few other constructors — notably `service/project` (registeredAt/archive), which looks like a sibling display bug worth a separate issue; (2) the fix is write-time only, so sessions stamped before this lands keep their local-zone timestamps.

harshitsinghbhandari (approved):
Approve

PR #217, opened 2026-06-12
fix(frontend): exclude nested node_modules from the vitest run

### LINKED ISSUE #216: vitest exclude misses nested node_modules (src/landing) — runs vendored third-party tests
## Bug

The renderer vitest config sets a custom `exclude` that **replaces** vitest's default `**/node_modules/**` with a root-anchored `node_modules/**`, so nested `node_modules` are no longer excluded. `frontend/src/landing` is a tracked, self-contained Next.js preview app with its own dependencies, so once its deps are installed, `npm test` scans `src/landing/node_modules/**` and runs **vendored third-party test suites** (zod, next, react-medium-image-zoom, style-to-js).

## Reproduction

```bash
cd frontend && npm install
cd src/landing && npm install        # install the landing preview's deps
cd .. && npm test
# → 8 failed test files / 20 failed tests, all under src/landing/node_modules/**
#   (e.g. src/landing/node_modules/zod/src/v4/classic/tests/codec-examples.test.ts)
```

## Root Cause

`frontend/vite.renderer.config.ts`:
```js
exclude: ["node_modules/**", "dist/**", "dist-electron/**", "e2e/**"],
```
Providing `test.exclude` overrides vitest's default (it does not merge). The default is `**/node_modules/**` (matches at any depth); `node_modules/**` is root-anchored and misses `src/landing/node_modules/**`.

## Fix

Use `**/node_modules/**` so nested `node_modules` (the landing preview's) stay excluded:
```js
exclude: ["**/node_modules/**", "dist/**", "dist-electron/**", "e2e/**"],
```

## Impact

A contributor who runs the landing preview (which the repo supports) and then `npm test` gets 20+ failures from code that isn't ours, masking real signal. CI is unaffected today (it doesn't install the landing's deps), so this is local-dev friction. Labels: `bug`, `priority: low`.

codebanditssss:
Closes #216.

## Problem

`frontend/vite.renderer.config.ts` set `test.exclude` to `["node_modules/**", ...]`. Providing `exclude` *replaces* vitest's default `**/node_modules/**` (it doesn't merge), and the root-anchored `node_modules/**` doesn't match nested `node_modules`. `src/landing` is a tracked, self-contained Next.js preview app with its own deps, so once those are installed, `npm test` collected and ran vendored third-party suites (zod, next, react-medium-image-zoom, style-to-js) → 20+ failures from code that isn't ours.

## Change

`"node_modules/**"` → `"**/node_modules/**"` so nested `node_modules` are excluded at any depth.

## Test

- Before: `npm test` collected 185 files / 2033 tests with 8 failed files / 20 failed tests, all under `src/landing/node_modules/**`.
- After: `npm test` collects only the renderer's **10 files / 121 tests, all passing**, with zero `src/landing/node_modules` references — and runs in ~6s instead of ~50s.

### DIFF
--- a/frontend/vite.renderer.config.ts
+++ b/frontend/vite.renderer.config.ts
@@ -74,7 +74,11 @@ export default defineConfig({
 	],
 	test: {
 		environment: "jsdom",
-		exclude: ["node_modules/**", "dist/**", "dist-electron/**", "e2e/**"],
+		// Anchor node_modules at any depth: a bare "node_modules/**" replaces
+		// vitest's default "**/node_modules/**" and only matches the root, so the
+		// tracked src/landing preview app's nested node_modules would otherwise
+		// have its vendored third-party test suites collected and run.
+		exclude: ["**/node_modules/**", "dist/**", "dist-electron/**", "e2e/**"],
 		globals: true,
 		setupFiles: "./src/renderer/test/setup.ts",
 	},


harshitsinghbhandari:
Reviewed locally. **Approve** — correct, minimal fix. Confirmed the root cause (supplying `test.exclude` replaces vitest's default, and slash-anchored `node_modules/**` only matches the root), that `src/landing` is tracked with un-committed nested `node_modules`, and that the single config covers both vitest entry points (`vitest.config.ts` re-exports it). No real test is dropped — all 10 suites live under `src/renderer`/`src/shared`.

harshitsinghbhandari (approved):
Approve — correct, minimal fix; verified locally.

PR #219, opened 2026-06-12
feat(frontend): remove a project from the sidebar

### LINKED ISSUE #218: Frontend: no way to remove a project from the UI (daemon/CLI already support it)
## Gap

The daemon and CLI both support removing a project (`DELETE /api/v1/projects/{id}` / `ao project rm <id>`), but the Electron renderer has **no UI to remove a project** — there's no action anywhere in the sidebar or project view. Once a project is added, the only way to remove it is the CLI.

Found while click-testing the app for the launch.

## Proposal

Surface project removal in the sidebar. Each project row already fades its session count on hover (`group-hover/menu-item:opacity-0`), leaving room for a hover action. Add a kebab (`SidebarMenuAction showOnHover`) that opens a small menu with **Project settings** and a destructive **Remove project**, wired to `DELETE /api/v1/projects/{id}`, optimistically removing the row and falling back to Home if the removed project was the active route.

(`ao project rm` only **archives** the registration — the repo on disk is untouched and it can be re-added — so this is safe/reversible.)

## Acceptance

- A "Remove project" action is reachable from each project row in the sidebar.
- Removing updates the sidebar immediately and navigates Home if that project was open.
- Backend/daemon unchanged (endpoint already exists).

codebanditssss:
Closes #218.

## Problem

The daemon and CLI both support removing a project (`DELETE /api/v1/projects/{id}` / `ao project rm`), but the renderer had **no UI** for it — once added, a project could only be removed from the CLI. Surfaced while click-testing the app.

## Change

- **Sidebar:** each project row gets a kebab (`SidebarMenuAction` with `showOnHover`, an absolute-positioned sibling of the row button — no nested-button issue). It reveals on hover where the session count fades, and opens a menu with **Project settings** and a destructive **Remove project**.
- **`_shell.tsx`:** new `removeProject` callback calls `apiClient.DELETE("/api/v1/projects/{id}")`, then optimistically drops the row from the workspace cache. If the removed project is the active route, the sidebar navigates Home.

No backend changes — the endpoint already exists. `ao project rm` only archives the registration (repo on disk untouched, re-addable), so this is reversible.

## Test

- `Sidebar.test.tsx` (new): opens the row kebab and clicks **Remove project**, asserting `onRemoveProject` is called with the project id. (Router hooks stubbed, since the Sidebar reads selection from the router.)
- `npm run typecheck` clean; prettier clean; the renderer vitest suite passes.

## Notes

- No confirmation dialog: the app has no AlertDialog/toast primitive yet, and the action is reversible — so it's a two-step intent (open kebab → click the red item). A confirm can follow when a dialog primitive lands.
- On delete failure the row resyncs via the existing 15s workspace refetch (logged to console); a visible error needs a toast surface the app doesn't have yet.

### DIFF
--- a/frontend/src/renderer/components/Sidebar.test.tsx
+++ b/frontend/src/renderer/components/Sidebar.test.tsx
@@ -1,0 +1,46 @@
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { describe, expect, it, vi } from "vitest";
+
+// The Sidebar reads selection from the router; stub the three hooks it uses so
+// the component renders without a full RouterProvider.
+vi.mock("@tanstack/react-router", () => ({
+	useNavigate: () => vi.fn(),
+	useParams: () => ({}),
+	useRouterState: () => "/",
+}));
+
+import { TooltipProvider } from "./ui/tooltip";
+import { SidebarProvider } from "./ui/sidebar";
+import { Sidebar } from "./Sidebar";
+import type { WorkspaceSummary } from "../types/workspace";
+
+const workspaces: WorkspaceSummary[] = [{ id: "proj-1", name: "my-app", path: "/p", type: "main", sessions: [] }];
+
+function renderSidebar(onRemoveProject = vi.fn().mockResolvedValue(undefined)) {
+	render(
+		<TooltipProvider>
+			<SidebarProvider>
+				<Sidebar
+					daemonStatus={{ state: "running" }}
+					workspaces={workspaces}
+					onCreateProject={vi.fn().mockResolvedValue(undefined)}
+					onRemoveProject={onRemoveProject}
+				/>
+			</SidebarProvider>
+		</TooltipProvider>,
+	);
+	return onRemoveProject;
+}
+
+describe("Sidebar", () => {
+	it("removes a project from the row kebab menu", async () => {
+		const user = userEvent.setup();
+		const onRemoveProject = renderSidebar();
+
+		await user.click(screen.getByRole("button", { name: "Actions for my-app" }));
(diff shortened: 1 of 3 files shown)


harshitsinghbhandari:
Reviewed locally. **Approve with minor comments.** Endpoint is typed, no nested-button issue (the kebab is an absolute sibling of the row button), and the cache update is safe (it runs post-success, not truly optimistic). Nice detail: the "reversible" framing is actually *more* accurate than the OpenAPI text — `Service.Remove` only archives (disk untouched), so the no-confirm decision is reasonable.

Minor, non-blocking: (1) cosmetic count/kebab overlap when the menu is open and the pointer leaves the row; (2) removing a project with active sessions leaves them running (pre-existing backend behavior, worth a follow-up); (3) optional tests for the `goHome()`-on-active and error/catch branches.

harshitsinghbhandari (dismissed):
Approve — verified locally; comments left are minor and non-blocking.

harshitsinghbhandari:
Re-reviewed after main moved — **this looks superseded.** While it was open, #225 "restore remove-project functionality" landed on main and implements the same feature across the same three files, which is why it's now CONFLICTING. Main's version is also a superset: it adds a `window.confirm` step (the confirmation I flagged earlier), `window.alert` error surfacing, and an in-flight guard — none of which this branch has.

I don't think there's a delta worth rebasing here; suggest closing as superseded by #225. (My earlier approval predates #225 and no longer applies.)

harshitsinghbhandari:
Closing as superseded: #225 landed the same remove-project feature on main (with a confirm step, error surfacing, and an in-flight guard), which is why this branch now conflicts. No delta left to rebase. Thanks for the work here — the capability shipped.

PR #246, opened 2026-06-15
fix(review): serialize concurrent triggers per worker to stop reviewer double-spawn

### LINKED ISSUE #242: fix(review): concurrent Trigger double-spawns reviewers (no serialization, no unique constraint)
## Bug

Concurrent `Trigger` calls for the same worker session at the same head SHA double-spawn reviewers and insert duplicate `running` review runs. The idempotency guard is a read-then-write with no serialization and no backing DB constraint.

**Source:** Live code analysis (bug hunt) | **Reported by:** @codebanditssss | **Analyzed against:** `198b797`
**Confidence:** High — confirmed no unique index on `review_run(session_id, target_sha)` in migration 0012.

## Reproduction

1. Issue two near-simultaneous `POST /api/v1/sessions/{id}/reviews/trigger` for the same worker at the same head SHA.
2. Both pass the `GetReviewRunBySessionAndSHA` idempotency check (neither has inserted yet), both spawn the deterministic `review-<id>` zellij session, and both `InsertReviewRun` → two `running` runs for one commit.

## Root Cause

`backend/internal/review/review.go:121` (`Trigger`) checks for an existing run (line ~151) then, much later, calls `e.launcher.Spawn(...)` (~173) and `e.store.InsertReviewRun(...)` (~211) with no lock/singleflight/transaction between check and write. `reviewerHandleID(workerID)` is deterministic (`"review-"+workerID`, `launcher.go:62`), so both spawns target the same zellij session id. `backend/internal/storage/sqlite/migrations/0012_add_review_tables.sql` declares `UNIQUE` only on the parent `review.session_id`, not on `review_run(session_id, target_sha)`. The route `POST /sessions/{id}/reviews/trigger` (`controllers/reviews.go:67`) has no per-session locking.

## Fix

Serialize triggers per worker session (keyed mutex / `singleflight` in the engine) and add a partial UNIQUE index on `review_run(session_id, target_sha)` so a duplicate `InsertReviewRun` fails and the engine can fall back to the existing run.

## Impact

- Two reviewer processes spawn against the same handle (second errors or duplicates), and the DB holds two `running` runs for one commit, defeating the idempotency design and confusing the UI's reviewer handle/run list.

## Related

- [#140](https://github.com/aoagents/ReverbCode/issues/140) — LCM intent-driven notification service foundation (adjacent review/lifecycle surface)

codebanditssss:
Closes #242.

## Problem

`Engine.Trigger` is a read-then-write — idempotency check (`GetReviewRunBySessionAndSHA`) → reviewer spawn → `InsertReviewRun` — with no serialization and no backing DB constraint. Two near-simultaneous triggers for the same worker at the same head SHA both pass the idempotency check (neither has inserted yet), both spawn a reviewer against the same deterministic `review-<id>` handle, and both insert a `running` run for one commit — defeating the idempotency design and leaving the UI's reviewer handle / run list inconsistent.

## Fix

Two layers:

1. **Per-worker keyed mutex** (`lockWorker`) held across the whole `Trigger` body. Concurrent triggers for the same worker now serialize; the loser re-reads the freshly-recorded run and short-circuits to `Created: false` instead of spawning. Distinct workers never contend.
2. **Partial unique index** on `review_run(session_id, target_sha)` (migration `0013`) as a cross-restart safety net. Rows with an empty `target_sha` (head not yet observed) are **excluded** so they aren't blocked — the engine lock still serializes those.

## Test

`TestTriggerConcurrentSameWorkerSpawnsOnce` fires 8 simultaneous triggers for one worker and asserts exactly one spawn, one recorded run, and one `Created: true` result. (Race-safe by construction: all shared access happens inside the serialized `Trigger` body; CI's `build-test` `-race` job validates — no local C compiler here.)

Gate: `go build ./...`, `gofmt -l`, `golangci-lint run` (0 issues), full `go test ./...`, and `internal/storage/sqlite` tests (which apply migration 0013) — all green.

### DIFF
--- a/backend/internal/domain/review.go
+++ b/backend/internal/domain/review.go
@@ -1,6 +1,15 @@
 package domain
 
-import "time"
+import (
+	"errors"
+	"time"
+)
+
+// ErrDuplicateReviewRun is returned by InsertReviewRun when a run already exists
+// for the same worker session and target commit (the partial unique index from
+// migration 0013). It lets the review engine fall back to the recorded run
+// instead of surfacing a raw storage error after a reviewer may have launched.
+var ErrDuplicateReviewRun = errors.New("domain: review run already exists for session and target sha")
 
 // Review is the per-worker code-review record: one row per worker session
 // (SessionID is unique). A repeat trigger reuses this row; the per-pass facts
--- a/backend/internal/review/review.go
+++ b/backend/internal/review/review.go
@@ -12,6 +12,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/google/uuid"
@@ -74,6 +75,12 @@ type Engine struct {
 	launcher Launcher
 	clock    func() time.Time
 	newID    func() string
+
+	// triggerMu guards triggerLocks; triggerLocks holds one mutex per worker
+	// session so concurrent Trigger calls for the same worker serialise (see
+	// lockWorker). Distinct workers never contend.
+	triggerMu    sync.Mutex
+	triggerLocks map[domain.SessionID]*sync.Mutex
 }
 
 // New wires an Engine from its dependencies, defaulting the clock and id source.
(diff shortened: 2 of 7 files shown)


Vaibhaav-Tiwari:
I’d like two changes before merge:

  1. Make migration 0013 safe for existing duplicate data.
     If a user already hit #242, their DB may already have multiple review_run rows with the same session_id +
     target_sha. CREATE UNIQUE INDEX will fail on that data and can stop daemon startup. The migration should
     dedupe those rows before creating the index, probably keeping the newest or most useful run and deleting/
     marking the others.

  2. Handle unique-constraint fallback in Trigger.
     If InsertReviewRun hits the new unique constraint anyway, Trigger should re-read GetReviewRunBySessionAndSHA
     and return that existing run with Created: false, instead of surfacing an error after the reviewer may
     already have launched.

PR #276, opened 2026-06-17
 feat: add agent catalog/auth API and safer orchestrator switching

nikhilachale:
## Summary

  This PR adds a daemon-backed agent catalog, exposes installed/authorized agent state to the frontend, and uses that data in project settings so users can choose worker/orchestrator agents more safely.

  It also adds orchestrator replacement handling: when the saved orchestrator agent changes, AO starts the replacement first and  only retires the previous orchestrator after the new one is up, so a failed replacement does not cause downtime.

  A key caveat is that agent-auth/login flows can interfere with replacement startup. If switching agents triggers the agent’s own bootstrap path, the replacement may come up outside AO’s normal orchestrator initialization path and miss the AO orchestrator system prompt.

  ## What Changed

  ### Backend

  - Added agent inventory service for:
      - supported agents
      - installed agents
      - authorized agents
      - counts for each

  - Added optional AgentAuthChecker capability on adapters.
  - Added shared CLI auth probing helper for adapters with cheap local auth checks.
  - Added GET /api/v1/agents.
  - Extended registry inventory entries to carry adapter manifest metadata for user-facing labels.
  - Added orchestrator replacement flow in the session service:
      - spawn replacement first
      - retire previous orchestrator only after successful replacement
      - preserve previous orchestrator when replacement startup fails

  - Added backend tests for agent catalog, controller responses, session replacement behavior, and related project/service wiring.

  ### Frontend

  - Regenerated API types for the new agents endpoint/DTOs.
  - Updated ProjectSettingsForm to:
      - load agent catalog from the daemon
      - show authorized agent options
      - handle installed-but-not-authorized states
      - surface orchestrator replacement pending state
      - allow retry once replacement is safe to perform

  - Added/updated tests for the new settings behavior.

  ## Why

  Before this change, the UI did not have a daemon-backed view of which agents are actually installed and authenticated on the local machine, and changing orchestrator agent config did not have a clear replacement flow.

  This PR makes agent selection more grounded in local runtime state and reduces the chance of downtime during orchestrator switches.

  ## Risks / Caveats

  - A large part of the file count comes from:
      - generated API artifacts
      - frontend tests
      - small per-adapter auth probe shims

  - The CLI/runtime/session model is unchanged outside the new inventory/auth and orchestrator replacement paths.
  - Generated files are included intentionally:
      - backend/internal/httpd/apispec/openapi.yaml
      - frontend/src/api/schema.ts

  - Auth/login flows remain a review risk. If switching agents triggers the agent’s own login/bootstrap flow, that flow can spawn  a fresh native session outside AO’s normal orchestrator startup path.

  - In that case, the replacement session may miss AO’s expected initialization, including the orchestrator system prompt.
  - The old orchestrator is intentionally preserved on replacement failure, but reviewers should pay close attention to whether
    replacement startup still guarantees AO system-prompt delivery.
    
**Closes** [#275](https://github.com/aoagents/ReverbCode/issues/275)


nikhilachale:
https://gist.githack.com/nikhilachale/3bfbe183e12be120b53ee37082304ffa/raw/agent-switcher-design-doc.html

backend/internal/session_manager/manager.go
```diff
@@ -437,6 +444,54 @@ func (m *Manager) Kill(ctx context.Context, id domain.SessionID) (bool, error) {
...
+	if !ok || rec.IsTerminated {
+		return nil
+	}
+	handle := runtimeHandle(rec.Metadata)
+	ws := workspaceInfo(rec)
+	if handle.ID == "" || ws.Path == "" {
+		return fmt.Errorf("retire %s: %w", id, ErrIncompleteHandle)
+	}
+	for attempt := 0; attempt < orchestratorRetireAttempts; attempt++ {
+		if err := m.runtime.Destroy(ctx, handle); err != nil {
```
neversettle17-101, line 484: This cutover path can leave the old orchestrator in a misleading state. `RetireOrchestrator` destroys the runtime before the session is marked terminated; if `Destroy` succeeds but a later `IsAlive`, workspace cleanup, or `MarkTerminated` step fails, the API reports `ORCHESTRATOR_REPLACEMENT_INCOMPLETE` and the UI tells users the previous orchestrator was kept alive, while the old process may already be gone and the DB row still looks active. For resiliency/UX, either mark terminal intent once destroy is known to have happened, or return a distinct recovery state that does not claim the previous orchestrator was preserved.

backend/internal/adapters/agent/codex/codex.go
```diff
@@ -217,6 +248,13 @@ func ResolveCodexBinary(ctx context.Context) (string, error) {
 	return "", fmt.Errorf("codex: %w", ports.ErrAgentBinaryNotFound)
 }
 
+func nvmNodeBinCandidates(home, binary string) []string {
+	matches, err := filepath.Glob(filepath.Join(home, ".nvm", "versions", "node", "*", "bin", binary))
+	if err != nil || len(matches) == 0 {
+		return nil
+	}
+	sort.Sort(sort.Reverse(sort.StringSlice(matches)))
+	return matches
```
neversettle17-101, line 257: \`nvmNodeBinCandidates\` is missing its closing brace — the function body never terminates before \`func resolveNativeWindowsCodex(...)\` starts on the next line, which nests a named function declaration inside another function. This is not valid Go and **the package will not compile** as currently pushed:

\`\`\`go
func nvmNodeBinCandidates(home, binary string) []string {
	matches, err := filepath.Glob(filepath.Join(home, ".nvm", "versions", "node", "*", "bin", binary))
	if err != nil || len(matches) == 0 {
		return nil
	}
	sort.Sort(sort.Reverse(sort.StringSlice(matches)))
	return matches
func resolveNativeWindowsCodex(path string) string {  // <-- missing `}` above
```

Needs a `}` after `return matches` to close `nvmNodeBinCandidates` before `resolveNativeWindowsCodex` starts. Given `TestResolveCodexBinaryFindsNVMInstallWhenPathIsSparse` was added in this same PR, this file couldn't have been built/tested in its current state — worth double-checking the push matches what was actually tested locally.

File: backend/internal/service/agent/service.go [hunk 3]
```diff
@@ -0,0 +1,106 @@
...
+
+// Inventory describes all daemon-supported agents and which are runnable here.
+type Inventory struct {
+	Supported  []Info `json:"supported"`
+	Installed  []Info `json:"installed"`
+	Authorized []Info `json:"authorized"`
+}
+
+// Service reports supported and locally runnable agent adapters.
+type Service struct {
```
neversettle17-101, line 27: Naming nit: every other `internal/service/*` package names its main file after either the package itself (`review.go` in package `review`) or its primary type (`manager.go` ↔ `Manager`, `action_service.go` ↔ `ActionService`). Here the file is `catalog.go` but the type is `Service`, not `Catalog` — so the file/type pairing doesn't match the convention elsewhere in `internal/service/`.

Since this is a brand-new single-file package, either renaming this file to `service.go` (matching `project`/`session`/`review`'s "main file" convention) or renaming the type to `Catalog` (to match this file) would read more consistently with the rest of the tree. Not blocking, just a naming consistency nit.

neversettle17-101 (changes requested):

backend/internal/service/session/service.go
```diff
@@ -269,23 +270,47 @@ func telemetryErrorKind(kind apierr.Kind) string {
...
+		var err error
+		existing, err = s.List(ctx, ListFilter{ProjectID: projectID, Active: &active, OrchestratorOnly: true})
 		if err != nil {
 			return domain.Session{}, err
 		}
-		for _, orch := range existing {
-			if _, err := s.Kill(ctx, orch.ID); err != nil {
-				return domain.Session{}, err
+	}
+	sess, err := s.Spawn(ctx, ports.SpawnConfig{ProjectID: projectID, Kind: domain.KindOrchestrator})
```
neversettle17-101, line 284: No locking around read-existing → spawn-new → retire-old in `SpawnOrchestrator(clean=true)`. Two concurrent calls for the same project will both snapshot the same `existing` orchestrators, both spawn a new one, and both retire the same old one — leaving two new orchestrators alive at once, breaking the "only one live coordinator" invariant this flag exists to guarantee. Needs a per-project lock (or a DB-level uniqueness check) around this sequence.

backend/internal/session_manager/manager.go
```diff
@@ -437,6 +451,73 @@ func (m *Manager) Kill(ctx context.Context, id domain.SessionID) (bool, error) {
...
+			continue
+		}
+		if err := m.recordRetiredTermination(ctx, id); err != nil {
+			return err
+		}
+		alive, err := m.runtime.IsAlive(ctx, handle)
+		if err != nil {
+			m.logger.Warn("session manager: old orchestrator probe failed after runtime destroy",
+				"session", id, "err", err)
+		} else if alive {
```
neversettle17-101, line 497: If `IsAlive` still reports `true` after `recordRetiredTermination` already succeeded, the loop retries `Destroy`/`recordRetiredTermination` again, and on final failure returns `ErrSessionStillAlive` — but the DB already says terminated. Since `List(Active: true)` won't surface a terminated row, nothing will ever retry killing this runtime again: a zombie orchestrator can leak silently while the system believes it's fully retired. Don't re-record termination once it has already succeeded; surface a distinct "terminated but still alive" signal so recovery tooling can retry the kill.

neversettle17-101:
Please resolve merge conflicts and test the flow end to end locally as well.

nikhilachale:
@neversettle17-101 
  1. codex.go missing brace: fixed.
     There is now a } after return matches, so resolveNativeWindowsCodex is no longer nested.

  2. catalog.go naming nit: fixed by renaming the file.
     backend/internal/service/agent/catalog.go is deleted and the same package code now lives in
     backend/internal/service/agent/service.go, matching the Service type.

  3. SpawnOrchestrator(clean=true) race: fixed.
     backend/internal/service/session/service.go now has per-project orchestrator locks around the read-
     existing → spawn-new → retire-old sequence. Tests were added for same-project serialization and
     allowing different projects to proceed concurrently.

  4. Retired orchestrator zombie handling: fixed.
     backend/internal/session_manager/manager.go now returns ErrRetiredSessionStillAlive after
     termination is recorded but the runtime still appears alive, instead of retrying/re-recording and
     ending with the generic ErrSessionStillAlive. The session service maps that to
     ORCHESTRATOR_REPLACEMENT_RECOVERY_REQUIRED.
     
     
Screenshot 2026-06-22 at 12 35 55
Screenshot 2026-06-22 at 12 35 38
Screenshot 2026-06-22 at 12 35 33

nikhilachale:
other than that 
@whoisasx  has added CreateProjectAgentSheet  and  some changes in ui around worker and orchestrator agent so add some fixes over them 


 - CreateProjectAgentSheet no longer uses hardcoded AGENT_OPTIONS.
    It now loads the shared /api/v1/agents catalog via React Query and shows supported agents with Needs
    install / Needs auth labels. Only authorized agents are selectable.

  - ProjectSettingsForm was refactored around the same agent catalog.
    It shows agent labels instead of raw IDs, supports manual Reload agents, warns for configured-but-
    missing or unauthorized agents, and shows an “Agent login needed” prompt when agents are installed
    but none are authorized.

  - Project settings validation was tightened.
    Existing projects missing worker/orchestrator role config now show Worker and orchestrator agents
    are required. and do not save until those are selected.


  - Sidebar project creation tests were updated.
    Tests now seed/mock the agent catalog and verify the project creation dialog handles async catalog
    loading.

  - NewTaskDialog.test.tsx timeout was increased to 10_000ms.
    That is just test stability for async typing/submission.

  - SessionsBoard.tsx cleanup.
    Removed duplicate/unused imports. No real behavior change there.

  - Select component styling changed.
    The popper viewport no longer forces h-[var(--radix-select-trigger-height)]; it uses trigger width
    without forcing dropdown height, so long agent option menus render correctly.

  - ShellLayout got daemon readiness handling.
    Before project creation it refreshes daemon status and errors clearly if the daemon is not ready. It
    also invalidates workspace queries when the daemon port changes.
    receives DefaultHarness: domain.AgentHarness(cfg.Agent) in projectsvc.Deps.

backend/internal/service/agent/catalog_test.go
neversettle17-101, line 1: nit: can we keep the file name service_test according to other prs?

backend/internal/service/session/service.go
```diff
@@ -257,23 +261,73 @@ func (s *Service) emitSpawnFailed(cfg ports.SpawnConfig, err error, durationMs i
...
 
 // SpawnOrchestrator spawns an orchestrator session for a project. When clean is
-// true it first tears down any active orchestrator(s) for that project so the new
-// one is the only live coordinator — a business rule that belongs here, not in the
-// HTTP controller.
+// true it performs a replacement cutover: start the new orchestrator first,
+// then retire any older active orchestrators for that project so a failed
+// replacement never causes downtime. This business rule belongs here, not in
+// the HTTP controller.
 func (s *Service) SpawnOrchestrator(ctx context.Context, projectID domain.ProjectID, clean bool) (domain.Session, error) {
```
neversettle17-101, line 268: when the new orchestrator is spawning what does the user experience look like?

Issue #251, opened 2026-06-16
bug: session inspector in summary section doesn't show PR details even though its been raised


neversettle17-101:
## Summary
Session Inspector Summary tab shows outdated information after agent completes work:
- **PR section**: Shows \"No pull request opened yet\" even though PR was created
- **Board (projects view)** kanban card: Shows \"no PR yet\" instead of PR number/state
- **/prs page**: Always empty, never shows any sessions with PRs

**Reported by:** Aditi Chauhan | **Session:** agent-orchestrator-2, PR #278 created but not reflected in UI
**Analyzed against:** `96d1649`
**Confidence:** High

## Reproduction
1. Spawn session with agent
2. Agent raises a PR and completes work
3. Open Session Inspector > Summary tab
4. **Expected:** Summary shows PR number/state/CI/mergeability/review; Board shows PR badge; /prs page lists the session
5. **Actual:** Summary shows \"No pull request opened yet\"; Board shows \"no PR yet\"; /prs page is empty

## Root Cause

`WorkspaceSession.pullRequest` (frontend/src/renderer/types/workspace.ts) is declared as an optional field (`pullRequest?: { number, state }`) but **never populated** anywhere in the codebase. The only place that builds `WorkspaceSession` objects is `useWorkspaceQuery.ts`, and it maps session fields manually without ever setting `pullRequest`.

Three places in the frontend read this dead field:
1. **SessionInspector.tsx (SummaryView, line 121)** — `const hasPr = Boolean(session.pullRequest);` gates whether the PR facts query runs. Since `session.pullRequest` is always undefined, `hasPr` is always false, the query never runs, and the summary always shows \"No pull request opened yet.\"
2. **SessionsBoard.tsx (SessionCard, line 223)** — `{session.pullRequest ? \`PR #${session.pullRequest.number} · ${session.pullRequest.state}\` : \"no PR yet\"}` for the kanban card PR badge. Same dead field, same problem.
3. **PullRequestsPage.tsx (line ~41)** — filters sessions by `Boolean(s.pullRequest)`, so the entire /prs page is always empty since nothing ever populates the field.

The backend already has the data: the `pr` table stores each session's PR, and `GET /api/v1/sessions/{sessionId}/pr` returns it correctly. The issue is purely that the frontend's session mapping never fetches/attaches this data.

## Fix
[...]

PR #230, opened 2026-06-14
feat(session): support multiple PRs per session

Pritom14:
## Summary

Lets a single session own several pull requests (a root plus stacked children) instead of being capped at one. The SQLite schema was already 1-session→many-PR (`pr.url` PK, `session_id` a plain FK with `idx_pr_session`, no UNIQUE on `session_id`), so this is a behavioural change across the **observe → persist → derive → react** pipeline, not a migration.

## State & user-action flow

The session's display status is **never stored** — it is derived at read time from durable PR facts. One full loop:

1. **Agent opens PR(s).** A root PR on the session branch (`feat/x`), optionally stacked children on the `feat/x/...` convention.
2. **OBSERVE** (`observe/scm`): each poll lists the repo's open PRs, attributes every PR whose source branch equals the session branch or descends from it to the owning session (longest-prefix wins), and persists the `pr` row including the source/target branch pair.
3. **DERIVE** (`service/session/status.go`, read-time): session status = **worst-wins aggregate** over the open PRs, by severity `ci_failed > changes_requested > draft > review_pending > pr_open > approved > mergeable`. A stacked child whose parent is still open is **excluded** from the aggregate (it cannot merge until the parent does). The stack model + per-PR facts are surfaced as `prs[]` on every session DTO, so the dashboard shows each PR's state.
4. **ACT** (`lifecycle/reactions.go`): per-PR nudges to the agent — CI failing → "fix CI" (with log tail); review changes/unresolved comments → "address feedback"; merge conflict → "rebase", **but only the bottom of the stack**; a stacked child blocked by an open parent is suppressed. Dedup signature is persisted per PR so a daemon restart never re-fires.
5. **User action closes the loop.** From the derived status + `prs[]` the user/agent fixes CI, addresses review, or rebases and pushes; the next poll re-observes and re-derives. No status is ever written back — only PR facts are.
6. **Completion.** The session terminates (→ status `merged`) only when **no PR is open and at least one merged**. A single PR merging while another stays open keeps the session alive.

### What changed, by layer

- **Observe** (`observe/scm/observer.go`): one refresh subject per open tracked PR; `discoverNewPRs` + `matchSession` do the branch-prefix attribution (longest match wins so a child session claims its own stacked PRs).
- **Derive** (`service/session/stack.go`, `status.go`): worst-wins aggregation + stack model (*B is a child of A iff `B.target_branch == A.source_branch` and A is open*), surfaced as `prs[]`.
[...]

PR #278, opened 2026-06-17
feat(review): enforce reviewer read-only via tool allowlist (#260)

### LINKED ISSUE #260: enhancement(review): enforce reviewer read-only via a tool allowlist instead of bypassPermissions + prompt
## Context

[#258](https://github.com/aoagents/ReverbCode/issues/258) / [PR #259](https://github.com/aoagents/ReverbCode/pull/259) make the reviewer run headless by launching it with `--permission-mode bypassPermissions`. That fixes the stall (the reviewer can now run `gh`/`ao` without a human approving each tool call), but it means the reviewer's read-only guarantee is enforced **only by the prompt** ("Do not push commits, edit files, or modify the branch — review only"), not by a sandbox. With bypass, the agent *could* technically edit files or push if the model misbehaved.

This issue tracks hardening that to an enforced read-only sandbox.

## The core constraint

`bypassPermissions` (≡ `--dangerously-skip-permissions`) skips the permission system entirely, so `--allowedTools`/`--disallowedTools` have **no effect** under it. To make an allowlist actually bite, the reviewer must launch in a **non-bypass** mode where allow/deny rules are honored — while still not prompting for the tools it legitimately needs (so it stays headless).

## Proposed approach

1. Add a tool-restriction field to `ports.LaunchConfig` (e.g. `AllowedTools []string` / `DisallowedTools []string`) and plumb it through the claude-code adapter's `GetLaunchCommand` to `--allowedTools` / `--disallowedTools`.
2. Launch the reviewer **off** `bypassPermissions`, with an allowlist scoped to exactly what a reviewer needs:
   - `Read`, `Grep`, `Glob`
   - `Bash(gh:*)`, `Bash(git diff:*)`, `Bash(git log:*)`, `Bash(git show:*)`, `Bash(ao review submit:*)`
   - and disallow `Edit` / `NotebookEdit` (and `Bash(git push:*)`, `Bash(git commit:*)`).
3. Resolve the `review.md` write: the reviewer is told to write `review.md` then `ao review submit --body review.md`. A no-`Write` policy breaks that, so either allow a narrowly-scoped `Write`, or change the flow so the body is passed without writing a file.

## Tradeoffs / open questions

- A tight allowlist risks the reviewer **stalling** if it reaches for an unlisted tool (e.g. a `cat`/`ls`/`rg` via `Bash` during exploration). The allowlist needs to be permissive enough for normal review work but tight enough to block writes — needs tuning/testing.
- Confirm current Claude Code semantics for headless TUI panes: which mode honors allow/deny lists without prompting on allowlisted tools, and how unlisted tools behave (deny vs prompt).
- Make the restriction reviewer-specific so worker sessions are unaffected.

## Related
- [#258](https://github.com/aoagents/ReverbCode/issues/258) — reviewer flow fixes (introduces bypassPermissions)
- [PR #259](https://github.com/aoagents/ReverbCode/pull/259) — the fix PR

neversettle17-101:
Fixes #260

## Summary

The reviewer's read-only guarantee was enforced **only by the prompt**. This makes it an enforced sandbox by launching the reviewer off `bypassPermissions` (which skips the permission system and ignores allow/deny rules) with a scoped tool allowlist.

1. **`ports.LaunchConfig`** gains `AllowedTools []string` / `DisallowedTools []string`. Empty = unrestricted, so worker sessions are unaffected.
2. **claude-code agent adapter** plumbs them to `--allowedTools` / `--disallowedTools`. Each list is comma-joined into one value so a rule containing spaces (e.g. `Bash(git diff:*)`) is not split into separate tool names.
3. **Reviewer adapter** launches off bypass with:
   - allow: `Read`, `Grep`, `Glob`, `Bash(gh:*)`, `Bash(git diff:*)`, `Bash(git log:*)`, `Bash(git show:*)`, `Bash(git status:*)`, `Bash(ao review submit:*)`
   - deny (defense in depth): `Edit`, `Write`, `NotebookEdit`, `Bash(git push:*)`, `Bash(git commit:*)`
4. **`review.md` write flow** resolved without granting `Write`: `ao review submit` gains `--body-text` for inline Markdown (`--body <path>` stays for humans; passing both is a usage error), and the prompt records the verdict inline instead of writing a file. No write tool is granted at all — a true read-only guarantee.

## Test

- `go build ./...`, full `go test ./...`, `gofmt`, `go vet` — all pass.
- New tests: agent-adapter flag emission (joined values; omitted when unset), a hermetic reviewer test (off-bypass + correct allow/deny lists, no `claude` binary needed), CLI `--body-text` inline + mutual-exclusion error.

## Notes

- **Accepted stall risk** (flagged in #260): the reviewer pane is interactive and headless. Off bypass, a tool that is neither allowed nor denied *prompts* — with no one to answer, it stalls. The scoped Bash allowlist means an unlisted bash command (`cat`, `rg`, `ls`) would stall. This is the tuning tradeoff the issue calls out; worth an end-to-end review run to confirm the allowlist is permissive enough.


backend/internal/adapters/reviewer/claudecode/claudecode.go
```diff
@@ -39,6 +70,11 @@ func (r *Reviewer) ReviewCommand(ctx context.Context, inv ports.ReviewInvocation
 		WorkspacePath: inv.WorkspacePath,
 		Prompt:        inv.Prompt,
 		SystemPrompt:  inv.SystemPrompt,
+		// Launch off bypassPermissions so the allow/deny lists are enforced:
+		// the reviewer's read-only guarantee is the sandbox, not just the
+		// prompt. Permissions is left unset (default mode) on purpose.
+		AllowedTools:    reviewerAllowedTools,
```
neversettle17-101, line 84: Permissions is left as the zero value here, which (per GetLaunchCommand) emits no --permission-mode flag and defers to the user's ~/.claude/settings.json defaultMode. If that default is bypassPermissions, this allow/deny list is silently ignored — the exact failure mode #260 wanted closed. Pass an explicit non-bypass mode (e.g. ports.PermissionModeAcceptEdits) instead of relying on ambient config.

neversettle17-101 (commented):
This closes the gap from #260 (allow/deny lists are wired through and the `review.md` write is avoided), but the central guarantee — "launch off bypassPermissions" — isn't actually enforced, so the fix can silently no-op back to the exact vulnerability #260 describes.

**Core issue:** `ReviewCommand` leaves `Permissions` as the zero value and doesn't set `Config` at all. In `GetLaunchCommand`, an empty `Permissions` (and empty `Config.Permissions`) normalizes to `PermissionModeDefault`, which emits **no** `--permission-mode` flag — Claude's TUI then resolves the starting mode from the user's own `~/.claude/settings.json` `defaultMode`. If that ambient default is `bypassPermissions` (plausible for AO operators who run agents autonomously), the reviewer launches in bypass mode, and per the adapter's own doc comment, "bypassPermissions ignores both lists" — so the allowlist/denylist do nothing and the reviewer is back to prompt-only enforcement, with no visible signal that the sandbox didn't take effect.

The new test (`TestReviewCommandLaunchesReadOnlyOffBypass`) doesn't catch this: it only asserts `agent.got.Permissions != PermissionModeBypassPermissions`, which is trivially true for the empty string, but doesn't verify the *resolved* launch is non-bypass.

To actually guarantee non-bypass (matching what #260 asked for: "the reviewer must launch in a non-bypass mode"), `ReviewCommand` should pass an explicit non-bypass mode (e.g. `ports.PermissionModeAcceptEdits` or `ports.PermissionModeAuto`) rather than relying on whatever `defaultMode` happens to be configured on the host. That makes the read-only guarantee independent of ambient user config, rather than a hope.

Everything else — the tool-flag plumbing, comma-joining to avoid splitting `Bash(git diff:*)`-style rules, the `--body-text` CLI flow, and the frontend reviewer-pane wiring — looks correct and well-tested.

neversettle17-101 (commented):
**This review is an approval.** (Posted as a COMMENT-event review, not APPROVE, because GitHub does not allow the PR author to approve their own PR.)

This addresses the finding from my previous review (PR #278 at commit 39906b3) directly and correctly.

**Fix verified:** `ReviewCommand` now sets `Permissions: ports.PermissionModeAuto` explicitly, instead of leaving it unset. Per `appendPermissionFlags`, this emits `--permission-mode auto`, which guarantees the launch is non-bypass regardless of what the host's `~/.claude/settings.json` `defaultMode` happens to be — closing the gap where an ambient `bypassPermissions` default would have silently disabled the allow/deny enforcement.

The test was updated in lockstep: `TestReviewCommandLaunchesReadOnlyOffBypass` now asserts `agent.got.Permissions == ports.PermissionModeAuto` instead of the previous `!= bypassPermissions` check, which is the right strengthening (the old assertion was trivially satisfied by the empty string that caused the bug).

`go build ./...` and the reviewer/agent adapter test suites pass. No further findings — approving.

harshitsinghbhandari (approved):
LGTM

neversettle17-101 (commented):
**This review is an approval.** (Posted as a COMMENT-event review, not APPROVE, because GitHub does not allow the PR author to approve their own PR.)

This head (`6372208`) is a merge of `origin/main` into the PR branch, with conflicts in `backend/internal/adapters/reviewer/claudecode/claudecode.go` and `backend/internal/review/prompt.go` (both touched by upstream #259, "reviewer posts to GitHub and records its verdict autonomously").

**Conflict resolution verified correct:**
- `claudecode.go`: the merge kept this PR's fix — `Permissions: ports.PermissionModeAuto` plus the allow/deny tool lists — rather than reverting to #259's `bypassPermissions` launch. The non-bypass sandbox guarantee from my earlier review is intact.
- `prompt.go`: the merge correctly combined #259's "post to GitHub first, then record with AO" step ordering with this PR's "no review.md write, use `--body-text`" flow. Step numbering (1/2/3) and the "only if step 2 fails, still run step 3" fallback line are consistent and don't reference removed steps.
- No leftover conflict markers (`grep` for `<<<<<<<`/`=======`/`>>>>>>>` across `backend/` is clean).

Diffing this PR's own files between the previously-reviewed head (`283f3e3`) and this one shows no new changes from this PR's own work — the rest of the diff (`frontend/src/api/schema.ts`, `pr-hydration.test.tsx`, `useWorkspaceQuery.*`) comes from the merged-in upstream commits (#251/#279, #230), not from this PR.

`go build ./...`, `go vet ./...`, `gofmt -l .`, and the full `go test ./...` backend suite all pass clean.

No further findings — approving.

Vaibhaav-Tiwari (changes requested):
i don't see any code level blocker, there are merge conflicts in the frontend files, just sync with the main and it can be merged

Vaibhaav-Tiwari (commented):
there's an issue that the PR claims to enforce a true read-only reviewer sandbox, but it launches Claude Code in auto permission mode. In that mode, the new --allowedTools list auto approves the listed tools, but it does not necessarily block every unlisted action; Claude can still auto-approve other safe local operations, including shell-based file writes inside the project. Since the deny list only blocks Edit, Write, NotebookEdit, git push, and git commit, the reviewer may still mutate the checkout through other Bash(...) commands. So the implementation improves the prompt-level restriction, but it does not actually provide the hard read-only guarantee described by the PR.

neversettle17-101:
### Heads-up: this doesn't fully guarantee a read-only reviewer

**The claim:** the reviewer can't modify your code.
**Reality:** it still can, through the shell.

The deny list blocks 5 things — `Edit`, `Write`, `NotebookEdit`, `git push`, `git commit` — but the reviewer can also run Bash, and there are many ways to change a file with a shell command that aren't on that list:

- `echo "x" > file`, `sed -i`, `rm`, `mv`, `cp`
- `python -c "open('f','w')..."`
- `git reset`, `git checkout`, `git stash`

A deny list can never cover every case — there's always one more command it missed. So this hardens the prompt-level restriction, but it isn't a hard read-only guarantee.

**One mitigating detail:** the reviewer runs in a pane with no human to approve prompts, so an unlisted write command would most likely *stall* rather than silently run. Not catastrophic — but still not the guarantee described.

**Suggested path:**
For V2 Follow-up: for a real guarantee, give the reviewer a read-only copy of the checkout at the OS level (read-only mount / container). Then no shell command can mutate the source, regardless of what the model runs.

The fix is *not* to keep extending the deny list — that chases an infinite list and still leaks.

backend/internal/adapters/reviewer/claudecode/claudecode.go
```diff
@@ -39,10 +70,12 @@ func (r *Reviewer) ReviewCommand(ctx context.Context, inv ports.ReviewInvocation
...
 		Prompt:        inv.Prompt,
 		SystemPrompt:  inv.SystemPrompt,
-		// The reviewer runs headless with no human to approve tool prompts; it
-		// is read-only by prompt and must run gh/ao on its own, so bypass the
-		// permission gate rather than stall on the first prompt.
-		Permissions: ports.PermissionModeBypassPermissions,
+		// Launch off bypassPermissions so the allow/deny lists are enforced.
+		// Set an explicit non-bypass mode instead of deferring to the user's
+		// Claude defaultMode, which may itself be bypassPermissions.
+		Permissions:     ports.PermissionModeAuto,
```
neversettle17-101: `PermissionModeAuto` makes the read-only guarantee classifier-dependent rather than allowlist-enforced: `--allowedTools` is additive auto-approval, not an exclusive whitelist, and the denylist only blocks Edit/Write/NotebookEdit/git push/git commit. Plain `Bash` (e.g. `> file`, `sed -i`, `rm`) is not denied, so a filesystem-mutating shell command is gated only by the auto classifier — not blocked. `PermissionModeDefault` would make it hard (unlisted tools prompt → stall headless, never execute). Either switch to default mode or document the residual Bash risk.

backend/internal/adapters/reviewer/claudecode/claudecode.go
```diff
@@ -31,6 +31,37 @@ func (r *Reviewer) Harness() domain.ReviewerHarness {
 
 var _ ports.Reviewer = (*Reviewer)(nil)
 
+// reviewerAllowedTools is the read-only tool allowlist the reviewer launches
+// with. The reviewer runs headless (no human to approve prompts) but must stay
+// read-only, so instead of bypassPermissions — which skips the permission
+// system entirely and ignores allow/deny rules — it launches in the default
```
neversettle17-101: Comment says it "launches in the default mode" but the code sets `PermissionModeAuto` (and there is a literal `PermissionModeDefault`). The comment describes default-mode semantics while the code uses auto — please reconcile.

backend/internal/review/prompt.go
```diff
@@ -21,14 +21,15 @@ Post your review on the pull request using the available review tooling (request
...
+1. You are read-only: you can read the checkout and run gh, git diff/log/show, and ao. You cannot write or edit files, so do not write a review file.
+2. Post your review on the pull request with `+"`gh`"+`, with inline comments for specific findings:
    - If changes are needed, request changes.
-   - If it is ready, approve it. GitHub does not let you approve a PR you opened — if the approval is rejected because you are the PR author, post the same review as a regular comment instead (a COMMENT-event review whose body states it is an approval).
-2. Write your full review to review.md and record the result with AO by running exactly:
+   - If it is ready, approve it. GitHub does not let you approve a PR you opened; if the approval is rejected because you are the PR author, post the same review as a regular comment instead (a COMMENT-event review whose body states it is an approval).
+3. Record the result with AO by passing your full review inline:
 
-    ao review submit --session %s --run %s --verdict <approved|changes_requested> --body review.md
+    ao review submit --session %s --run %s --verdict <approved|changes_requested> --body-text "<your full review as Markdown>"
```
neversettle17-101: Inline `--body-text "<...>"` is shell-fragile: review Markdown commonly contains `"`, backticks, `$`, and `\`, which break double-quoted shell args or trigger expansion — and this is the step that must survive even when the gh post fails. A stdin form (`--body -` from a heredoc) avoids both worktree writes and shell-mangling. At minimum recommend single-quoting.

neversettle17-101 (commented):
**Requesting changes.**

Good, well-tested change — the plumbing of `AllowedTools`/`DisallowedTools` through `LaunchConfig`, the comma-join for space-bearing rules, the `--body-text`/`--body` mutual-exclusion, and the hermetic reviewer test are all clean. Two things should be reconsidered before merge, both centered on whether this actually delivers the *enforced* read-only guarantee the PR (and #260) promises.

### 1. `auto` mode does not enforce read-only for Bash — the guarantee is classifier-dependent, not allowlist-enforced

The reviewer launches in `PermissionModeAuto`. In claude-code, `--allowedTools` is *additive auto-approval*, not an exclusive whitelist — tools that are neither allowed nor denied fall through to the permission mode. In `auto` mode a safety classifier auto-approves whatever it deems "safe." `--disallowedTools` only hard-blocks `Edit`/`Write`/`NotebookEdit`/`git push`/`git commit`, but plain `Bash` is not denied, so a filesystem-mutating shell command (`> file`, `tee`, `sed -i`, `rm`, `git stash`, etc.) is gated only by the classifier's judgment, not by the allowlist. That undercuts the PR's "true read-only guarantee" / "enforced sandbox" framing — the worker's checkout can still be mutated via Bash if the classifier approves.

`PermissionModeDefault` would make the guarantee hard: every non-allowlisted tool *prompts*, and headless that *stalls* rather than executes, so nothing outside the allow list can ever run. The tradeoff is more stalls (`cat`/`ls`/`rg` would stall instead of being classifier-approved) — but that is exactly the read-only-hardness vs. stall-avoidance tradeoff, and the PR is sold on hardness. Please either switch to `default` mode, or scope `Bash` itself down, or explicitly document that the read-only guarantee for Bash is classifier-dependent residual risk rather than allowlist-enforced.

Relatedly, the `reviewerAllowedTools` doc comment says it "launches in the **default mode** where these rules are honored" — but the code sets `auto`, and there is a literal `PermissionModeDefault`. The comment describes default-mode semantics while the code does something else; please reconcile.

### 2. Inline `--body-text "<...>"` in the prompt is shell-fragile

The prompt instructs the headless reviewer to run `ao review submit ... --body-text "<your full review as Markdown>"`. Review bodies routinely contain double quotes, backticks, `$`, and backslashes — inside a double-quoted shell argument those break the quoting or trigger expansion, so the verdict-recording step (the one that must survive even when posting fails) is the most likely thing to mangle or fail. A stdin path (`--body -` reading from a heredoc) would avoid worktree writes *and* be robust to arbitrary Markdown. At minimum, recommend single-quoting in the prompt guidance.

### Minor / acknowledged
- The accepted stall risk for unlisted Bash is noted in the PR; with `auto` mode it is largely mitigated, but that mitigation is the same mechanism that weakens finding #1 — worth being explicit about which behavior you are relying on.

backend/internal/adapters/reviewer/claudecode/claudecode.go
neversettle17-101: Addressed in c8b70172 (re: review 4577706991). Switched the reviewer from \`PermissionModeAuto\` to \`PermissionModeDefault\`. In default mode the allowlist auto-approves only the read-only tools the reviewer needs; every other tool — including a filesystem-mutating \`Bash\` (\`> file\`, \`sed -i\`, \`rm\`, \`git stash\`) — now prompts and, with no human on the headless pane, stalls rather than executing. Read-only is enforced by the allowlist instead of left to the auto classifier. The remaining stall risk for unlisted Bash is exactly the tradeoff called out, and is now the relied-upon behavior.

backend/internal/adapters/reviewer/claudecode/claudecode.go
neversettle17-101: Addressed in c8b70172 (re: review 4577706991). Reconciled: the code now sets \`PermissionModeDefault\`, matching the comment, which I also rewrote to state explicitly why default (not auto, not bypass) is required — auto would let a classifier-"safe" shell write through, default cannot. The test now asserts \`PermissionModeDefault\`.

backend/internal/review/prompt.go
neversettle17-101: Addressed in c8b70172 (re: review 4577706991). Replaced the inline \`--body-text "<...>"\` step with a quoted heredoc piped to \`ao review submit --body -\`: the prompt now instructs the reviewer to pass the review on stdin so \`"\`, backticks, \`$\`, and \`\\\` pass through literally with no shell mangling and no worktree write. Added \`--body -\` stdin support to the submit command, with a test that round-trips Markdown containing exactly those characters.

PR #289, opened 2026-06-17
fix(project): resolve default branch from origin/HEAD, not checked-out branch

### LINKED ISSUE #286: bug(project): `ao project add` permanently adopts whatever branch is checked out at registration time as the project default, diverging from main
## Bug
`ao project add` records the repo's **currently checked-out branch** as the project's permanent `defaultBranch`, instead of the repo's actual default branch (`origin/HEAD`, typically `main`). If a user happens to be mid-feature-work (e.g. on `fix/some-branch`) when they register a project, AO mistakes that personal/transient branch for the project's base branch — and every worker spawned afterward bases its worktree off that stale feature branch instead of `main`, forever, with no resync and no warning.

**Source:** Discovered live while triaging [#284](https://github.com/aoagents/ReverbCode/issues/284)/[#285](https://github.com/aoagents/ReverbCode/issues/285) — the `agent-orchestrator` project's stored config shows `"defaultBranch": "fix/pr-attachment"`, a personal scratch branch, not `main`.
**Analyzed against:** `7c9ae53`
**Confidence:** High — traced the exact code path and reproduced the live DB state.

## Reproduction
1. Check out a non-default branch in a registered repo's working directory (e.g. `git checkout -b fix/something`).
2. Run `ao project add --path <repo>` while that branch is checked out.
3. `project.Config.DefaultBranch` is persisted as `fix/something`.
4. Switch back to `main` in the working directory — AO's stored config is unaffected.
5. Spawn any worker: its worktree bases off `fix/something`, not `main`, with no indication anything is wrong.

Confirmed live: `~/.ao/data/ao.db` → `projects.config` for project `agent-orchestrator` contains `{"defaultBranch":"fix/pr-attachment", ...}` — `fix/pr-attachment` was simply the branch checked out in the main working tree at the moment the project was registered, not the repo's actual default branch (`origin/HEAD` → `main`).

## Root Cause
`backend/internal/service/project/service.go`:
- Line 202-206 (in `Add`): `if row.Config.DefaultBranch == "" { if branch := resolveDefaultBranch(path); ... { row.Config.DefaultBranch = branch } }`
- Line 254-260: `resolveDefaultBranch` runs `git -C path symbolic-ref --short HEAD` — i.e. **whatever branch happens to be checked out right now**, not the repo's actual default branch.

This was introduced by [PR #209](https://github.com/aoagents/ReverbCode/pull/209) (closing [#208](https://github.com/aoagents/ReverbCode/issues/208)), which fixed a real problem — repos on `master`/`develop`/`trunk` previously fell back to a hardcoded `main` and failed every spawn with `BRANCH_NOT_FETCHED`. The fix's approach (snapshot `symbolic-ref --short HEAD` once, at `project add` time) conflates two different things:
- **The repo's actual default/trunk branch** (what `origin/HEAD` resolves to — stable, shared, intended target for new worktrees)
- **Whatever branch happens to be checked out locally right now** (transient, personal, could be any in-progress feature branch)

For a repo cloned fresh and never checked out elsewhere, these coincide, which is presumably why #209's tests passed. But for any already-active local clone (the common case for someone setting up AO on an existing project), they diverge — and once persisted, `Config.DefaultBranch` is `""`-gated (line 202), so it's a one-time snapshot that never self-corrects even after the user returns to `main`.

## Fix (options, not prescriptive)
- Prefer the repo's actual remote default branch: `git -C path symbolic-ref --short refs/remotes/origin/HEAD` (falling back to the current `symbolic-ref --short HEAD` approach only if no `origin/HEAD` is set, e.g. no remote configured) — this targets the repo's trunk rather than the registrar's transient checkout.
- Surface the detected `defaultBranch` in the `ao project add` output / project settings UI so a user notices if it's wrong, with an easy `ao project set-config --default-branch main` escape hatch (which already exists per `docs/cli/README.md`).
- Consider re-validating/re-resolving on a project's `ao project add` re-run or via `ao project set-config`, rather than only ever setting it once at first registration.

## Impact
- Every worker session spawned in an affected project silently bases its work off a stale personal branch instead of `main` — diffs, PRs, and merges all target/diverge from the wrong base, with no error or warning anywhere in the flow.
- Affects any project registered while the registrar had a non-default branch checked out — likely common, since `ao project add` is typically run against an already-active local clone, not a fresh checkout.

## Related
- [#208](https://github.com/aoagents/ReverbCode/issues/208) — the original bug this regressed from (closed)
- [#284](https://github.com/aoagents/ReverbCode/issues/284) — PR auto-attach disconnection (separate but adjacent: both are "session/project branch tracking goes stale and never resyncs" bugs)

codebanditssss:
Closes https://github.com/aoagents/ReverbCode/issues/286 
## What

`project add` was detecting the default branch with `git symbolic-ref --short HEAD`, which returns the *currently checked-out* branch — not the repo's actual default. Adding a project while sitting on a feature branch (e.g. `fix/pr-attachment`) persisted that feature branch as the project default, so every new session worktree based off it instead of `main`.

## Fix

`resolveDefaultBranch` now reads the remote default first:

1. `git symbolic-ref --short refs/remotes/origin/HEAD` → the real default (strip `origin/`)
2. fall back to `symbolic-ref --short HEAD` only when there's no remote default
3. empty on error → caller falls back to `DefaultBranchName` (`main`)

This keeps the original behaviour that motivated the detection — repos whose default is `master`/`develop` are still recorded correctly — while no longer latching onto the active feature branch.

## Tests

- `TestManager_AddPrefersOriginHeadOverCheckedOutBranch` — HEAD on `fix/pr-attachment`, origin/HEAD = `main` → records `main`
- `TestManager_AddPrefersOriginHeadNonMain` — origin/HEAD = `master`, HEAD on feature branch → records `master`
- existing `TestManager_AddDetectsNonMainDefaultBranch` still passes

all green via `go test ./internal/service/project/`.

### DIFF
--- a/backend/internal/service/project/service.go
+++ b/backend/internal/service/project/service.go
@@ -246,12 +246,25 @@ func resolveGitOriginURL(path string) string {
 	return strings.TrimSpace(string(out))
 }
 
-// resolveDefaultBranch returns the repo's currently checked-out branch via
-// `git -C path symbolic-ref --short HEAD`. A detached HEAD, missing repo, or any
-// other git error returns an empty string — `project add` must not fail just
-// because the branch can't be resolved (the caller falls back to
-// DefaultBranchName).
+// resolveDefaultBranch returns the repo's default branch, preferring the
+// remote's default (`origin/HEAD`) over the currently checked-out branch. This
+// matters because the user may have the repo on a feature branch when adding the
+// project: keying off HEAD would persist that feature branch as the project
+// default and base every session worktree on it. `origin/HEAD` reflects the
+// real default (e.g. `master`, `develop`) regardless of the active branch.
+//
+// Falls back to the checked-out branch when origin/HEAD is unset (no remote, or
+// it was never fetched). A detached HEAD, missing repo, or any other git error
+// returns an empty string — `project add` must not fail just because the branch
+// can't be resolved (the caller falls back to DefaultBranchName).
 func resolveDefaultBranch(path string) string {
+	if out, err := exec.Command(
+		"git", "-C", path, "symbolic-ref", "--short", "refs/remotes/origin/HEAD",
+	).Output(); err == nil {
+		if ref := strings.TrimSpace(string(out)); ref != "" {
+			return strings.TrimPrefix(ref, "origin/")
+		}
+	}
 	out, err := exec.Command("git", "-C", path, "symbolic-ref", "--short", "HEAD").Output()
 	if err != nil {
 		return ""
--- a/backend/internal/httpd/controllers/sessions_test.go
+++ b/backend/internal/httpd/controllers/sessions_test.go
@@ -23,6 +23,7 @@ type fakeSessionService struct {
 	cleanupProjects []domain.ProjectID
 	cleanupResult   []domain.SessionID
 	cleanupSkipped  []sessionsvc.CleanupSkipped
+	spawnErr        error
 	claimErr        error
 	listPRErr       error
 }
@@ -51,6 +52,9 @@ func (f *fakeSessionService) List(_ context.Context, filter sessionsvc.ListFilte
 }
 
 func (f *fakeSessionService) Spawn(_ context.Context, cfg ports.SpawnConfig) (domain.Session, error) {
+	if f.spawnErr != nil {
+		return domain.Session{}, f.spawnErr
+	}
 	now := time.Now().UTC()
 	s := domain.Session{SessionRecord: domain.SessionRecord{ID: domain.SessionID(string(cfg.ProjectID) + "-2"), ProjectID: cfg.ProjectID, IssueID: cfg.IssueID, Kind: cfg.Kind, Harness: cfg.Harness, Activity: domain.Activity{State: domain.ActivityIdle, LastActivityAt: now}, CreatedAt: now, UpdatedAt: now}, Status: domain.StatusIdle}
 	f.sessions[s.ID] = s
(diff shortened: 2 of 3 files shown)


neversettle17-101:
Can you also test it end to end apart from unit tests & add the testing details here?

backend/internal/service/project/service.go
```diff
@@ -246,12 +246,25 @@ func resolveGitOriginURL(path string) string {
...
+	if out, err := exec.Command(
+		"git", "-C", path, "symbolic-ref", "--short", "refs/remotes/origin/HEAD",
+	).Output(); err == nil {
+		if ref := strings.TrimSpace(string(out)); ref != "" {
+			return strings.TrimPrefix(ref, "origin/")
+		}
+	}
 	out, err := exec.Command("git", "-C", path, "symbolic-ref", "--short", "HEAD").Output()
 	if err != nil {
 		return ""
```
neversettle17-101, line 270: If we are not able to fetch a branch make sure we return correct error.

neversettle17-101 (changes requested):

codebanditssss:
Done. I added API-level end-to-end coverage for the branch-not-fetched case.

If the requested branch cannot be resolved/fetched, the workspace layer returns `ErrWorkspaceBranchNotFetched`, the service maps it to `BRANCH_NOT_FETCHED`, and `POST /api/v1/sessions` now has coverage verifying it returns HTTP 400 with that code.

Testing run:
- `go test ./internal/adapters/workspace/gitworktree`
- `go test ./internal/session_manager`
- `go test ./internal/service/session`
- `go test ./internal/httpd/controllers`

neversettle17-101 (approved):
Please test it also E2E once merged.

PR #307, opened 2026-06-18
Feat/backend telemetry v0

laxmanclo:
Implemented end-to-end telemetry for the rewrite on feat/backend-telemetry-v0.

This adds local SQLite telemetry plus optional PostHog export in the Go daemon, wires core backend events (daemon.started, cli.invoked, usage errors, spawn success/failure, 5xx, panic, project/onboarding, waiting-input dwell), and adds renderer-side PostHog analytics/error capture in Electron. The desktop app now ships with org-owned PostHog defaults, shares one stable install identity across backend + renderer, and emits a canonical ao.app.active event so DAU/WAU can be charted cleanly in PostHog.


laxmanclo:
@AgentWrapper can you review this PR?

Vaibhaav-Tiwari (changes requested):
- High: Raw errors are exported to PostHog without a central sanitizer.
    The PostHog sink hashes project/session IDs, but then copies all payload fields directly into remote properties in backend/internal/adapters/telemetry/posthog.go:148-150.
    Several call sites put raw errors into payloads, for example HTTP 5xx errors in backend/internal/httpd/log.go:55-64, session spawn failures in backend/internal/service/
    session/service.go:236-240, and CLI usage errors in backend/internal/httpd/router.go:189-193. Renderer telemetry also captures global exceptions and unhandled rejections in
    frontend/src/renderer/lib/telemetry.ts:25-37, plus React component stacks in frontend/src/renderer/components/TelemetryBoundary.tsx:19-23. These strings can contain local
    paths, command args, repo names, branch names, URLs, or daemon error details. The remote sink should enforce an allowlist/redaction layer instead of trusting every event
    producer.

  - Medium: AO_TELEMETRY_METRICS=on currently enables event capture.
    Config documents separate event and metric toggles in backend/internal/config/config.go:118-119, but backend/internal/daemon/telemetry_wiring.go:13-16 creates the event sink
    whenever either Events or Metrics is true. Since this PR only implements event-style payloads, a user enabling metrics can accidentally persist/export events too. The wiring
    should only create event sinks when event capture is enabled, or introduce a separate metrics sink.

PR #309, opened 2026-06-18
fix(session): make worker branch namespace child-safe

### LINKED ISSUE #308: AO session branch namespace blocks child PR branches
## Bug
AO tells worker sessions to create PR branches under their session branch namespace, for example `ao/agent-orchestrator-7/fix-terminal-clipboard-305`, but the session root branch itself is created as `ao/agent-orchestrator-7`. Git stores branch refs path-like, so an existing `refs/heads/ao/agent-orchestrator-7` prevents creating `refs/heads/ao/agent-orchestrator-7/<topic>`.

Observed in session `agent-orchestrator-7`: the agent attempted:

```bash
git checkout -b ao/agent-orchestrator-7/fix-terminal-clipboard-305
```

Git failed with:

```text
fatal: cannot lock ref 'refs/heads/ao/agent-orchestrator-7/fix-terminal-clipboard-305': 'refs/heads/ao/agent-orchestrator-7' exists; cannot create 'refs/heads/ao/agent-orchestrator-7/fix-terminal-clipboard-305'
```

The agent then fell back to `fix-terminal-clipboard-305`, opened PR #306 from that branch, and AO did not attach PR facts to the session because the SCM observer only attributes PRs whose source branch equals the stored session branch or has it as a `/` child prefix.

**Analyzed against:** local main checkout on 2026-06-18
**Confidence:** High

## Reproduction
1. Spawn a worker session whose default branch is `ao/<session-id>`.
2. From that worktree, try to create `ao/<session-id>/<topic>`.
3. Git rejects the child branch because `refs/heads/ao/<session-id>` already exists.
4. If the agent creates a fallback branch outside the namespace and opens a PR, AO's observer does not attribute the PR to the session.

## Root Cause
- Worker branch defaults to `ao/<session-id>` in `backend/internal/session_manager/manager.go`.
- Worker instructions ask agents to use `<session-branch>/<topic>` for additional PRs.
- Git cannot have both `refs/heads/ao/<session-id>` and `refs/heads/ao/<session-id>/<topic>`.
- PR discovery in `backend/internal/observe/scm/observer.go` relies on exact or slash-child source branch matching, so fallback branches outside the namespace remain untracked.

## Fix
The fix helps by making AO's branch convention valid in Git and matching AO's PR discovery logic.

Before:

```text
session branch: ao/agent-orchestrator-7
expected child PR branch: ao/agent-orchestrator-7/fix-terminal-clipboard-305
```

Git rejects that because `refs/heads/ao/agent-orchestrator-7` already exists, so it cannot also create `refs/heads/ao/agent-orchestrator-7/fix-terminal-clipboard-305`.

After:

```text
session branch: ao/agent-orchestrator-7/root
sibling PR branch: ao/agent-orchestrator-7/fix-terminal-clipboard-305
```

Now Git can store both branches:

```text
refs/heads/ao/agent-orchestrator-7/root
refs/heads/ao/agent-orchestrator-7/fix-terminal-clipboard-305
```

AO also treats the parent namespace `ao/agent-orchestrator-7` as owned by the session, so GitHub PR heads under that namespace are attributed back to `agent-orchestrator-7`.

## Impact
- Worker sessions can create PRs successfully but AO may not discover or display them.
- Summary/activity tabs drift because PR facts are never inserted into SQLite.
- Agents following AO's own prompt hit a Git ref layout error.

neversettle17-101:
Fixes #308

## Summary
- Change default worker branches from `ao/<session-id>` to `ao/<session-id>/root` so sibling PR branches like `ao/<session-id>/<topic>` are valid Git refs.
- Teach SCM PR attribution that a `/root` session branch also owns sibling branches under the same namespace.
- Update worker branch instructions and CLI help text.

## Root Cause
AO created the session root branch as `ao/<session-id>` while instructing agents to create additional PR branches as `ao/<session-id>/<topic>`. Git cannot create both refs because `refs/heads/ao/<session-id>` blocks `refs/heads/ao/<session-id>/<topic>`.

## Tests
- `GOCACHE=/private/tmp/ao-go-build-cache go test ./internal/session_manager`
- `GOCACHE=/private/tmp/ao-go-build-cache go test ./internal/observe/scm`
- `GOCACHE=/private/tmp/ao-go-build-cache go test ./internal/cli -run 'TestSpawn'`

### DIFF
--- a/backend/internal/cli/root_test.go
+++ b/backend/internal/cli/root_test.go
@@ -230,6 +230,7 @@ func TestStopDoesNotShutdownUnverifiedReusedPID(t *testing.T) {
 func TestStopUsesShutdownEndpoint(t *testing.T) {
 	cfg := setConfigEnv(t)
 	shutdownCalled := make(chan struct{}, 1)
+	var shutdownSeen atomic.Bool
 	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		switch r.URL.Path {
 		case "/healthz":
@@ -240,6 +241,7 @@ func TestStopUsesShutdownEndpoint(t *testing.T) {
 			if err := runfile.Remove(cfg.runFile); err != nil {
 				t.Fatal(err)
 			}
+			shutdownSeen.Store(true)
 			shutdownCalled <- struct{}{}
 			_, _ = fmt.Fprintf(w, `{"status":"shutting_down","service":%q,"pid":%d}`, daemonmeta.ServiceName, os.Getpid())
 		default:
@@ -253,7 +255,12 @@ func TestStopUsesShutdownEndpoint(t *testing.T) {
 	}
 
 	out, _, err := executeCLI(t, Deps{
-		ProcessAlive: func(pid int) bool { return pid == os.Getpid() },
+		ProcessAlive: func(pid int) bool {
+			if pid != os.Getpid() {
+				return false
+			}
+			return !shutdownSeen.Load()
+		},
 	}, "stop", "--json")
 	if err != nil {
 		t.Fatal(err)
--- a/backend/internal/cli/spawn.go
+++ b/backend/internal/cli/spawn.go
@@ -121,7 +121,7 @@ func newSpawnCommand(ctx *commandContext) *cobra.Command {
 	})
 	f.StringVar(&opts.project, "project", "", "Project id to spawn the session in (required)")
 	f.StringVar(&opts.harness, "harness", "", "Agent harness / --agent: claude-code, codex, aider, opencode, grok, droid, amp, agy, crush, cursor, qwen, copilot, goose, auggie, continue, devin, cline, kimi, kiro, kilocode, vibe, pi, autohand (default: the daemon's AO_AGENT)")
-	f.StringVar(&opts.branch, "branch", "", "Branch for the session worktree (default: ao/<session-id>)")
+	f.StringVar(&opts.branch, "branch", "", "Branch for the session worktree (default: ao/<session-id>/root)")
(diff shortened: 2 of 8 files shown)


Pritom14:
@neversettle17-101 can you check this edge case : 

If two sessions in the same repo somehow end up with `ao/p-1` (no suffix) AND `ao/p-1/root`, both claim the prefix `ao/p-1` for sibling PRs and the longest-prefix tiebreak ties at length 6. The winner is then determined by iteration order of byRepo[repoKey] — nondeterministic. Realistic only if a user manually picks `ao/<id>` as a branch override while another defaulted session lives at `ao/<id>/root.`

neversettle17-101:
On the `ao/p-1` vs `ao/p-1/root` edge case: I do not think this will show up in many normal flows.

The ambiguous case requires all of these to exist in the same repo:

- Session A stored branch: `ao/p-1`
- Session B stored branch: `ao/p-1/root`
- PR head branch: `ao/p-1/something`

That is unlikely because existing sessions from before this fix may have `ao/p-1`, but newly spawned sessions after the fix will get new IDs like `ao/p-2/root`, `ao/p-3/root`, etc. AO should not naturally create both `ao/p-1` and `ao/p-1/root` for two different sessions. Also, if `ao/p-1` still exists as a Git ref, creating `ao/p-1/root` usually fails at Git ref creation anyway.

The more realistic compatibility case is an old running session with `session.branch = ao/p-1` trying to create `ao/p-1/topic`; Git will reject that because `ao/p-1` already exists as a leaf ref. That affects only sessions spawned in the short window after multi-PR support landed and before this fix. Those sessions can finish their root PR, or any fallback PR can be manually claimed if needed.

So I would keep this PR simple and not add extra tie-break logic unless we see this ambiguity in real state or a reviewer wants the deterministic fallback explicitly.

PR #310, opened 2026-06-18
feat(windows): ConPTY terminal, zellij discovery, agent launcher trampoline, codex shim resolution

miniMaddy:
## Summary

Four atomic commits adding Windows-native support across the daemon stack:

### 1. eat(terminal): Windows ConPTY support for /mux attach
- Adds go-pty v0.2.3 (ConPTY) and bumps golang.org/x/sys v0.44.0
- Implements pty_windows.go using ConPTY via winpty.New() with detach-grace + Kill fallback
- Plumbs nv param through PTYSource.AttachCommand on all platforms

### 2. eat(zellij): discover zellij binary on Windows and raise command timeout
- Probes zellij.exe via PATH then well-known install locations (LOCALAPPDATA, ProgramFiles)
- Raises default command timeout to 30 s on Windows (vs. the existing 10 s Unix default)

### 3. eat(zellij,cli): Windows agent launcher trampoline for codex argv
- New gentlaunch package: serialises agent argv to a temp JSON spec file, reads-and-removes on launch
- New hidden o launch subcommand: the trampoline entry point invoked inside the zellij pane
- Windows zellij Create writes a direct layout (no shell wrapper) and starts the pane via Start-Process (CREATE_NEW_CONSOLE, hidden window)
- PowerShell launch uses -EncodedCommand (UTF-16 LE base64) to avoid quoting issues

### 4. eat(codex): Windows binary resolution, terminal compat flags, TOML literal strings
- ResolveCodexBinary follows .cmd/.ps1 npm shims to the native .exe on Windows
- Appends --no-alt-screen on Windows to prevent alternate-screen escape sequences breaking the ConPTY pane
- TOML config values are now emitted as literal strings (single-quote) to avoid backslash escapes in Windows paths

## Testing
- All changed packages pass go test: 	erminal, zellij, gentlaunch, cli, codex
- go build ./... and go vet ./... clean on both Unix and Windows
- Pre-existing TestBuild_MatchesEmbedded failure in httpd/apispec/specgen is present on main too (stale embedded openapi.yaml) — not introduced by this PR


Priyanchew:
@miniMaddy ci failures resolve karlo

Vaibhaav-Tiwari (changes requested):
go test -v ./internal/adapters/runtime/zellij fails on Windows in TestRuntimeIntegration:

  Create: zellij runtime: list panes ao_itest_zj: exit status 1: There is no active session!

  This is directly in scope for the Windows issue. The new Windows zellij path starts the session fire-and-forget, then
  polls list-panes, but the integration session is already gone before discovery. That means the runtime path still has
  an unhandled Windows failure mode and the changed package does not pass its own Windows integration test.

  I also noticed a smaller cleanup issue: writeLayout creates an ao-launch-*.json spec on Windows, but Create only
  defers removal of the layout file. If zellij startup or pane discovery fails, the launch spec can be left behind in
  %TEMP%.

  Tests I ran:

  go test -v ./internal/terminal
  go test -v ./internal/daemon
  go test -v ./internal/cli
  go test -v ./internal/adapters/agent/codex

  Those passed.

  go test -v ./internal/adapters/runtime/zellij

  Failed as above.

Vaibhaav-Tiwari (changes requested):
The lint/doc cleanup looks fine, but the Windows runtime blocker is still present. On Windows:

  go test -v ./internal/adapters/runtime/zellij

  now fails in TestRuntimeIntegration at SendMessage:

  SendMessage: zellij runtime: paste message ao_itest_zj/terminal_0: exit status 1: There is no active session!

  So Create now returns, but the zellij session is gone before the runtime can paste/send input. That is still directly
  in scope for the Windows issue because AO needs a stable runtime handle after spawn, not just a successful Create
  return.

  Other focused tests passed:

  go test -v ./internal/terminal
  go test -v ./internal/adapters/agent/codex
  go test -v ./internal/daemon ./internal/cli

Priyanchew:
tested, it works

miniMaddy:
Works

PR #329, opened 2026-06-19
Feat/posthog error tracking redaction

laxmanclo:
This improves telemetry without widening what we send remotely.

Renderer exceptions now go into PostHog Error Tracking(before it was just tracking the error, not  capturing in depth, which gives us proper grouped UI issues. We also scrub local paths and local URLs before capture. On the backend, failures still emit as events, but now include better structured metadata like component, operation, error kind/code, and
fingerprints so they are easier to group and investigate without leaking raw internals. Backend tele can still be improved a lot, but requires lots of change.


Vaibhaav-Tiwari (changes requested):
Reviewed locally. I found one blocking privacy issue.

  The PR claims local URLs are redacted before PostHog capture, but embedded local URLs inside
  exception text are not redacted. For example, a message like:

  `failed to fetch http://127.0.0.1:3037/api/v1/projects?token=secret`

  would still be sent through unchanged. Same issue applies to embedded `localhost` and `app://
  renderer/...` URLs. The current logic only handles `file:///...` when embedded, or local URLs when
  the whole string is exactly the URL.

  Local checks:
  - Backend telemetry/http/session tests passed.
  - Frontend typecheck passed.
  - Frontend telemetry tests passed.
  - Full frontend suite mostly passed; one unrelated `NewTaskDialog` test timed out at default 5s
  but passed when rerun with a longer timeout.

  Recommendation: don’t merge yet. Please extend the redaction logic and tests for embedded
  `localhost`, `127.0.0.1`, `[::1]`, and `app://renderer` URLs in exception messages/stack fields.

Vaibhaav-Tiwari (approved):
LGTM

Issue #333, opened 2026-06-20
main is red: duplicate goose migration version 0014 panics every DB-backed test


codebanditssss:
## Bug

`main` is red because two SQLite migrations in `backend/internal/storage/sqlite/migrations/` share goose version `0014`. goose v3 derives a migration version from the numeric filename prefix and panics on a duplicate version.

The duplicate migrations are:

| File | Source | Notes |
| --- | --- | --- |
| `0014_review_run_retry_failed.sql` | [PR #318](https://github.com/aoagents/ReverbCode/pull/318) | Rewrites the `review_run` unique index |
| `0014_telemetry_events.sql` | [PR #307](https://github.com/aoagents/ReverbCode/pull/307) | Adds `telemetry_event` |

`backend/internal/storage/sqlite/db.go:96` runs `goose.Up()` from `Open()`, so the duplicate-version panic cascades through every package that opens a store.

## Symptom

CI panics in DB-backed tests. One visible failure is `TestE2E_StoreWriteToBroadcast`, which reports a panic referencing `migrations/0014_review_run_retry_failed.sql [recovered, repanicked]`.

Affected packages include `cdc`, `cli`, `daemon`, `httpd/controllers`, `integration`, `service/project`, `storage/sqlite`, and `storage/sqlite/store`.

`lint`, `api-drift`, and `gitleaks` stay green because they never open a DB. That is why PR checks passed before `main` was re-run.

[PR #323](https://github.com/aoagents/ReverbCode/pull/323) is not the cause. It touched no migration files and only re-ran CI against the already-broken `main`.

## Reproduction

```bash
cd backend
go test ./internal/storage/sqlite/...
```

## Fix

Rename the telemetry migration so goose sees a unique version:

```text
backend/internal/storage/sqlite/migrations/0014_telemetry_events.sql
->
backend/internal/storage/sqlite/migrations/0015_telemetry_events.sql
```

This is safe because `0014_telemetry_events.sql` only adds the `telemetry_event` table and has no dependency on `0014_review_run_retry_failed.sql`, which only rewrites the `review_run` unique index. goose reads the migration version from the filename prefix, so the rename is the whole fix. There is no embedded version constant to update.

## Impact

- `main` is fully red for every DB-backed test package.
- The daemon can panic on fresh-DB startup because `Open()` runs migrations via `goose.Up()`.

PR #336, opened 2026-06-20
test(storage): guard against duplicate goose migration version prefixes

neversettle17-101:
## Summary
- Adds a fast, dependency-free Go test that statically scans embedded migration filenames in `backend/internal/storage/sqlite/migrations` for duplicate numeric version prefixes (the digits before the first `_`, same as goose derives).
- Fails with a clear, actionable message (`duplicate migration version 0014: fileA vs fileB`) instead of relying on goose's runtime panic.
- Guards against a recurrence of #333, where two parallel PRs each independently added a `0014_*.sql` migration — neither conflicted individually in CI, but the collision broke every DB-backed test once both merged into main.

## Test plan
- [x] `gofmt -l backend` — clean
- [x] `go test ./internal/storage/sqlite/... -run TestMigrationVersionsAreUnique -v` passes against current main's clean migration set (0001–0015)
- [x] Temporarily added a scratch `0014_*.sql` file and confirmed the test fails with the expected message; removed it afterward
- [x] `go test -race ./internal/storage/sqlite/...` passes, confirming the existing `.github/workflows/go.yml` `build-test` job (`go test -race ./...`) picks this up with no workflow changes needed

### DIFF
--- a/backend/internal/storage/sqlite/migrate_unique_version_test.go
+++ b/backend/internal/storage/sqlite/migrate_unique_version_test.go
@@ -1,0 +1,64 @@
+package sqlite
+
+import (
+	"testing"
+)
+
+// TestMigrationVersionsAreUnique guards against #333: two parallel PRs each
+// added a migration file with the same numeric prefix (0014_review_run_retry_failed.sql
+// and 0014_telemetry_events.sql). Neither PR conflicted on its own, so CI on
+// each passed individually; the collision only surfaced after both merged,
+// when goose.Up() derives the version from that prefix and panics on the
+// duplicate. This test statically scans the embedded migration filenames for
+// a repeated version prefix, so the conflict is caught by `go test` with a
+// clear message instead of a goose panic at runtime.
+func TestMigrationVersionsAreUnique(t *testing.T) {
+	entries, err := migrationsFS.ReadDir("migrations")
+	if err != nil {
+		t.Fatalf("read migrations dir: %v", err)
+	}
+
+	seen := map[string]string{} // version prefix -> filename
+	for _, e := range entries {
+		name := e.Name()
+		if e.IsDir() {
+			continue
+		}
+
+		prefix := versionPrefix(name)
+		if prefix == "" {
+			t.Errorf("migration %q has no leading numeric version prefix that goose can parse", name)
+			continue
+		}
+
+		if other, dup := seen[prefix]; dup {
+			t.Errorf("duplicate migration version %s: %s vs %s", prefix, other, name)
+			continue
+		}
+		seen[prefix] = name
+	}
+}
+
+// versionPrefix returns the leading run of digits before the first
+// underscore in a migration filename — the same substring goose parses as
+// the migration's version number. It returns "" if the filename has no
+// underscore or no leading digits.
+func versionPrefix(filename string) string {
+	idx := -1
+	for i, c := range filename {
+		if c == '_' {
+			idx = i
+			break
+		}
+	}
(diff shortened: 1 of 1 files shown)


backend/internal/storage/sqlite/migrate_unique_version_test.go
```diff
@@ -0,0 +1,64 @@
...
+		}
+		seen[prefix] = name
+	}
+}
+
+// versionPrefix returns the leading run of digits before the first
+// underscore in a migration filename — the same substring goose parses as
+// the migration's version number. It returns "" if the filename has no
+// underscore or no leading digits.
+func versionPrefix(filename string) string {
```
neversettle17-101: \`versionPrefix\` dedupes by the raw string prefix, but goose derives the version with \`goose.NumericComponent\`, which parses that prefix via \`strconv.ParseInt\` and compares the resulting \`int64\` (pressly/goose v3 migration.go:363-379).

So this doesn't quite replicate goose's actual collision rule: two filenames whose prefixes are different strings but the same integer — e.g. \`014_foo.sql\` vs \`0014_bar.sql\`, or \`14_foo.sql\` vs \`0014_bar.sql\` — parse to the identical goose version and would panic at runtime, but this function returns them as distinct keys (\`\"014\"\` vs \`\"0014\"\`, etc.), so the test would pass and miss exactly the class of bug it's meant to catch.

Low risk today since all current migrations use a consistent 4-digit zero-padded convention, but the whole point of this guard is to catch convention drift before merge (same root cause as #333). Suggest dedup'ing on the parsed \`int64\` (via \`goose.NumericComponent\` or an equivalent \`strconv.ParseInt\`) instead of the raw string.

neversettle17-101 (commented):
**Review verdict: changes requested** (posted as a COMMENT-event review — GitHub does not allow requesting changes on your own PR).

Good, narrow fix for a real recurrence risk (#333/#334). Builds clean, gofmt clean, and `TestMigrationVersionsAreUnique` passes against the current migration set. One blocking finding inline: the dedup key doesn't exactly replicate how goose derives versions, which undercuts the test's stated purpose.

backend/internal/storage/sqlite/migrate_unique_version_test.go
```diff
@@ -0,0 +1,64 @@
+package sqlite
+
+import (
+	"testing"
+)
+
+// TestMigrationVersionsAreUnique guards against #333: two parallel PRs each
+// added a migration file with the same numeric prefix (0014_review_run_retry_failed.sql
+// and 0014_telemetry_events.sql). Neither PR conflicted on its own, so CI on
```
neversettle17-101: Remove the comments specific to any PR -- just add a concise method to do it.

neversettle17-101 (commented):
**Review verdict: approved** (posted as a COMMENT-event review — GitHub does not allow approving your own PR).

This revision fixes the blocking finding from my previous review: the test now calls `goose.NumericComponent` directly (the same function goose itself uses) and keys the dedup map on the parsed `int64` rather than a hand-rolled string-prefix parser. That means prefixes differing only in leading zeros/digit count (e.g. "014" vs "0014") are now correctly caught as collisions, matching goose's actual runtime behavior.

Verified: `go build ./...`, `go vet ./internal/storage/sqlite/...`, `gofmt -l` all clean, and `go test ./internal/storage/sqlite/... -run TestMigrationVersionsAreUnique -v` passes against the current migration set.

neversettle17-101:
Merging the PR-- this introduces tests to avoid migration conflict.

Issue #337, opened 2026-06-20
bug(review): AO-reviewer changes-requested verdict doesn't notifies the worker session


neversettle17-101:
## Bug
Reviewer inline comments aren't injected into the agent's PTY session — no "address review feedback" nudge is sent.

**Source:** PR [aoagents/ReverbCode#336](https://github.com/aoagents/ReverbCode/pull/336), session `agent-orchestrator-3`
**Confidence:** High

## Root Cause
`review.Engine.Submit` (`backend/internal/review/review.go:278-317`) only persists the AO reviewer's verdict/body (`UpdateReviewRunResult`) — it never notifies the worker. The comment on `Submit` says "AO does not post the review — the reviewer agent posts it to the PR itself," so on `VerdictChangesRequested` the worker only learns about it later via the generic SCM poll loop (`backend/internal/observe/scm/observer.go`'s `needsReviewRefresh`), which is unreliable: it's gated on GitHub's `reviewDecision` field, which never becomes `CHANGES_REQUESTED` for self-reviews or `COMMENT`-state reviews.

## Fix direction (not: poll GitHub harder)
Don't make the generic SCM-comment poller catch this — that would re-inject *every* PR comment from anyone into the worker session, which isn't what we want. Instead, fix it at the source: `Submit` already knows the verdict and body for the AO reviewer's own pass. When `verdict == VerdictChangesRequested`, have `Submit` (or its caller in `backend/internal/httpd/controllers/reviews.go`) directly message the worker session — same mechanism `lifecycle.Manager` uses (`ports.AgentMessenger.Send`, see `backend/internal/lifecycle/reactions.go:441`) — with the changes-requested body and an instruction to address it and reply on the GitHub thread. This is event-driven off the AO reviewer's own submit action, not a poll.

neversettle17-101:
**Superseded** — polling harder (like legacy AO's ETag-based refresh) isn't the right fix; it would re-inject every PR comment from anyone, not just the AO reviewer's verdict. See updated issue body for the correct fix: notify the worker directly from `review.Engine.Submit` when it records a changes-requested verdict.

PR #340, opened 2026-06-20
fix(review): notify worker on changes_requested instead of relying on SCM poll (#337)

neversettle17-101:
## What #337 required

`review.Engine.Submit` only persisted the reviewer's verdict/body via `UpdateReviewRunResult` — it never told the worker session anything. On a `changes_requested` verdict the worker was left to discover the feedback through the SCM poll loop (`observe/scm` `needsReviewRefresh`), which is gated on GitHub's `reviewDecision`. That decision never becomes `CHANGES_REQUESTED` for self-reviews or COMMENT-state reviews, so the nudge frequently never fired.

This PR makes it event-driven: when `verdict == VerdictChangesRequested`, `Submit` now messages the worker's live pane directly through `ports.AgentMessenger.Send` — the same mechanism `lifecycle.Manager` uses for its SCM nudges. The messenger is wired into the review engine in `daemon/lifecycle_wiring.go` (the per-daemon messenger already in scope). Approved verdicts do not message the worker. The body is reviewer-authored text pasted into a PTY, so it is run through `domain.SanitizeControlChars` first, matching the lifecycle reaction path.

We deliberately did **not** touch the generic SCM-comment poller — re-injecting every PR comment from anyone is out of scope.

## Extended flow (review-id round-trip)

So the worker knows *which* review to address:

1. The reviewer agent posts its review with `gh api --method POST repos/{owner}/{repo}/pulls/{number}/reviews ... --jq '.id'` and passes that id through a new `ao review submit --review-id` flag (prompt updated in `internal/review/prompt.go`). It posts via `gh api` rather than `gh pr review` because that is the only way to attach inline comments, and the POST response already carries the created review's id — so the id is captured from the call the reviewer makes anyway, exactly (no array-ordering, pagination, or read-back heuristics).
2. The daemon stores the id on the `review_run` row (new `github_review_id` column, **migration 0016** — latest before this was 0015) and includes it in the changes-requested message to the worker.
3. After addressing the feedback, the worker replies on the review referencing that id with what it changed, and resolves the inline review comment threads it addressed.

The id flows: CLI flag → `SubmitReviewInput.githubReviewId` → service → `Engine.Submit` → `UpdateReviewRunResult` (now also sets the column) and into the worker message. It is optional throughout: if the reviewer can't read the id back, everything still works without it (the reply/resolve instruction is simply omitted).

## AO internal review vs. external SCM review (worker message)

The worker now gets two clearly different nudges depending on origin:

- **External GitHub reviewer** (lifecycle SCM loop, unchanged): "A reviewer left feedback on your PR. Address it and push."
- **AO internal review** (`Engine.Submit`, this PR): identifies itself as an AO code reviewer (not an external GitHub PR reviewer), and — when the review id is known — asks the worker, after pushing the fix, to **reply on that review referencing its id** with what it changed and to **resolve the review comment threads it addressed**.

## "Resolve conversation" tradeoff

GitHub's `resolveReviewThread` (GraphQL) only applies to **inline review-comment threads tied to a line**, not to a top-level PR review object. The reviewer's prompt has it post **inline comments for specific findings** alongside the top-level request-changes verdict, so the per-finding threads *are* resolvable — the worker is told to resolve those after addressing them. The top-level review object itself is **not** resolvable, so for that we use option **(a)**: the worker replies on the review referencing the id (a plain comment) rather than claiming to "resolve" it.

I rejected the heavier option (b) — restructuring around `domain.ReviewRun` to track per-line comment threads and their resolution state in AO — because it fights the one-verdict/one-body-per-run model and isn't needed: the worker agent can run `resolveReviewThread` itself against the threads the reviewer already created.

## Tests

- `internal/review`: hermetic unit tests with a fake `AgentMessenger` — changes_requested messages the worker with body + review id, is identified as an AO internal review, and carries the reply/resolve instruction; the instruction is omitted when no id is supplied; approved does not message; messenger error is surfaced (happy-path-only, per scope). Existing Submit tests updated for the new signature.
- `internal/storage/sqlite/store`: round-trips `github_review_id` through insert/update/get.
- Regenerated `openapi.yaml` and `frontend/src/api/schema.ts` (code-first API contract); `gen/*` via sqlc.
- `go build ./...`, `go vet ./...` clean. Affected packages pass `go test -count=1`. (Pre-existing zellij runtime integration tests fail locally on IPC-socket-path-length / local-binary-path — environment-specific, unrelated.)

## Out of scope / follow-ups (deferred per issue)

- Resiliency: retry on message-send failure, dead/terminated worker session, double-submit idempotency. Today `Submit` surfaces a send error to the caller and does nothing fancier — happy path only.
- AO-side tracking of which threads were resolved, if richer per-finding state is wanted later.


backend/internal/review/prompt.go
```diff
@@ -21,14 +21,21 @@ Post your review on the pull request using the available review tooling (request
...
 
 Do these steps in order:
-1. Post your review on the pull request with `+"`gh`"+`, with inline comments for specific findings:
-   - If changes are needed, request changes.
-   - If it is ready, approve it. GitHub does not let you approve a PR you opened — if the approval is rejected because you are the PR author, post the same review as a regular comment instead (a COMMENT-event review whose body states it is an approval).
+1. Post your review on the pull request and capture its id in one call. Post with `+"`gh api`"+` rather than `+"`gh pr review`"+`: it is the only way to attach inline comments, and its response carries the created review's id, so AO can tell the worker exactly which review to address.
+
+    gh api --method POST repos/{owner}/{repo}/pulls/{number}/reviews \
+      -f event=REQUEST_CHANGES -f body="<summary>" \
+      -f 'comments[][path]=<file>' -F 'comments[][line]=<n>' -f 'comments[][body]=<finding>' \
```
neversettle17-101: `gh api` cannot build an array of objects from `-f`/`-F` flags. Only a key *ending in* `[]` appends (scalars); `comments[][path]` is treated as a literal field name, so this sends `{"comments[][path]": ...}` and GitHub rejects/ignores the inline comments. Since attaching inline comments via `gh api` is the stated reason for moving off `gh pr review`, this needs to use an `--input` JSON body, e.g.:

```
gh api --method POST repos/{owner}/{repo}/pulls/{number}/reviews --input - --jq '.id' <<'JSON'
{ "event": "REQUEST_CHANGES", "body": "<summary>",
  "comments": [ { "path": "<file>", "line": <n>, "body": "<finding>" } ] }
JSON
```

The `--jq '.id'` capture and the approve/COMMENT fallback prose are fine as-is.

neversettle17-101 (commented):
**Verdict: changes_requested** (posted as a COMMENT review — GitHub does not allow REQUEST_CHANGES on your own PR).

The core change (event-driven worker nudge on `changes_requested` + `github_review_id` round-trip) is solid, well-plumbed, and well-tested. Requesting one change to the new reviewer-agent prompt before merge — details below and inline.

## Blocking: the documented `gh api` command won't attach inline comments

`gh api`'s `-f`/`-F` flags build a *flat* body and only treat a key ending in `[]` as an array-append of scalars. `comments[][path]=...` is taken as a literal top-level field, not as "append an object to the `comments` array", so the body becomes `{"comments[][path]": ...}` rather than `{"comments":[{"path":...,"line":...,"body":...}]}`. GitHub will 422 / drop those, so the review posts with no inline comments — which is the whole reason the latest commits switched from `gh pr review` to `gh api`. Use `--input` with a JSON body instead (see inline comment).

## Non-blocking

If `messenger.Send` fails, `Submit` returns an error *after* `UpdateReviewRunResult` already marked the run `complete`; a retried `ao review submit` then fails the `status='running'` guard, leaving the run recorded but the worker un-nudged. Explicitly deferred (resiliency/idempotency) in the PR description, so acceptable — just flagging.

Everything else — migration 0016 numbering, sqlc gen, store round-trip, wiring, and the hermetic tests — looks good. `go build`, `go vet`, `gofmt`, and affected tests pass locally.

review.md
```diff
@@ -0,0 +1,42 @@
+# Review: fix(review): notify worker on changes_requested instead of relying on SCM poll (#337)
```
neversettle17-101: This file is a reviewer-agent scratch artifact (a review writeup of this PR) that was accidentally committed onto the worker branch in `5df20c9`. It's unrelated to the feature and should not be merged. Please `git rm review.md`, and consider adding it (or the reviewer's scratch path) to `.gitignore` so reviewer output can't be committed onto worker branches again.

neversettle17-101 (commented):
**Verdict: changes_requested** (posted as a COMMENT review — GitHub does not allow REQUEST_CHANGES on your own PR).

## Re-review at `fcc441e`

My earlier blocking finding is resolved — `prompt.go` now posts the review with an `--input -` JSON heredoc, so inline comments form a proper `"comments": [ {path, line, body} ]` array GitHub accepts, instead of the `comments[][path]` flat-field form `gh api` can't serialize. The approve/COMMENT fallback prose and `--jq '.id'` capture are intact. `go build` clean. The core feature remains solid and well-tested.

## Blocking: a stray `review.md` was committed into the PR

The PR now adds a 42-line `review.md` at the repo root (added in `5df20c9`, reformatted by `fcc441e`). Its contents are a code-review writeup of this very PR — a reviewer-agent scratch artifact that leaked onto the worker branch. It's unrelated to the feature and shouldn't merge. Please `git rm review.md`, and consider gitignoring the reviewer's scratch path so this can't recur. See inline comment.

## Non-blocking (deferred per PR description)

If `messenger.Send` fails, `Submit` returns an error after `UpdateReviewRunResult` already marked the run `complete`; a retried `ao review submit` then fails the `status='running'` guard. Explicitly deferred (resiliency/idempotency) — flagging only.

Once `review.md` is removed, this is good to merge.

neversettle17-101 (commented):
**Verdict: approved** (posted as a COMMENT review — GitHub does not allow the PR author's account to submit APPROVE).

## Re-review at `154b99e`

Both of my prior findings are fully resolved, with sensible extra hardening.

- **Stray `review.md` (was blocking) — fixed.** `40d3d19` removes it and `.gitignore` now ignores `/review.md`. Better still, the root cause is gone: `ao review submit --body -` reads the review from stdin (`io.ReadAll(cmd.InOrStdin())`) and the reviewer prompt passes it on a heredoc, so nothing is written into the checkout at all.
- **Submit error after DB commit (was non-blocking) — fixed.** `600ea1b` notifies the worker *before* `UpdateReviewRunResult`, so a failed `messenger.Send` leaves the run `running` and a retried submit runs again instead of tripping the `status='running'` guard. Trade-off documented in the comment (matches lifecycle `sendOnce`); new test asserts both stays-running and successful retry.
- **`event=COMMENT` always** (`237548f`/`154b99e`): correct — the reviewer posts from the PR author's own account, where GitHub rejects APPROVE and REQUEST_CHANGES. The verdict still reaches AO via step 2.
- Earlier `gh api` inline-comments fix (JSON `--input` body) is in place.

**Verification:** `go build ./...`, `go vet`, `gofmt` clean; `go test ./internal/review/... ./internal/cli/...` pass incl. new stdin-body + retry tests; diff vs `main` is focused, migration correctly numbered 0016, no stray files.

Nice work — ready to merge.

neversettle17-101 (commented):
**Verdict: approved** (posted as a COMMENT review — GitHub does not allow the PR author's account to submit APPROVE).

## Re-review at `154b99e` (unchanged since my last approval)

The branch is identical to the commit I reviewed and approved last round; re-confirming. Both of my earlier findings remain resolved:

- **Stray `review.md` (was blocking) — fixed.** Removed in `40d3d19`, `/review.md` gitignored, and the root cause is gone: `ao review submit --body -` reads the review from stdin and the reviewer prompt passes it on a heredoc, so nothing is written into the checkout.
- **Submit error after DB commit (was non-blocking) — fixed.** `600ea1b` notifies the worker *before* `UpdateReviewRunResult`, so a failed send leaves the run `running` and a retry works instead of tripping the `status='running'` guard. New test asserts stays-running + successful retry.
- **`event=COMMENT` always** correctly matches the PR-author-account constraint; the verdict still reaches AO via step 2. Earlier `gh api` inline-comments fix (JSON `--input` body) is in place.

**Verification:** `go build ./...`, `go vet`, `gofmt` clean; `go test ./internal/review/... ./internal/cli/...` pass incl. new stdin-body + retry tests; diff vs `main` is focused, migration correctly numbered 0016, no stray files.

Ready to merge.

neversettle17-101:
Tested E2E Flow
Screenshot 2026-06-21 at 6 11 21 PM
Screenshot 2026-06-21 at 6 11 29 PM

PR #344, opened 2026-06-20
fix(review): gate idempotency on a recorded verdict, not run status

### LINKED ISSUE #342: bug(review): Trigger's idempotency check treats a dead Running row as up-to-date, blocking re-review after the reviewer terminal is killed
## Bug
If a reviewer pass is killed mid-run (e.g. the user closes/cancels the reviewer's terminal directly instead of letting it finish), retrying review for the same commit returns "Review is already up to date for this commit." even though no review actually completed.

**Analyzed against:** `a8a3056`
**Confidence:** High — traced exact code path

## Root Cause
`Engine.Trigger` (`backend/internal/review/review.go:189-193`) decides idempotency purely from row existence + status, with no liveness check:

```go
if existing, ok, err := e.store.GetReviewRunBySessionAndSHA(ctx, workerID, targetSHA); err != nil {
    return TriggerResult{}, err
} else if ok && existing.Status != domain.ReviewRunFailed {
    return TriggerResult{Run: existing, ReviewerHandleID: review.ReviewerHandleID, Created: false}, nil
}
```

Any non-`Failed` row for `(session, targetSHA)` — including one still `Running` — short-circuits with `Created: false`, which the frontend renders as "Review is already up to date" (`frontend/src/renderer/components/SessionInspector.tsx:208`).

Nothing transitions a `Running` row to `Failed`/`Complete` when the underlying reviewer process dies outside of `Submit()`. `Submit` (review.go:278-317) only fires when the reviewer agent itself reports a verdict. Killing the terminal bypasses that entirely, so the row stays `Running` forever. The only place that checks actual liveness (`e.launcher.Alive`, review.go:247) is in the "reuse a live pane" branch further down — which is never reached because the idempotency check above already returned.

## Fix
In the idempotency check, when `existing.Status == domain.ReviewRunRunning`, call `e.launcher.Alive(ctx, review.ReviewerHandleID)` before short-circuiting. If the handle is dead, mark the run `Failed` (reuse the existing `failRun` pattern) and fall through to spawn a fresh review instead of returning the stale row.

## Impact
A worker session can get permanently stuck unable to re-review a commit once a reviewer pass is killed mid-run, with no workaround short of pushing a new commit.

neversettle17-101:
Fixes aoagents/ReverbCode#342

## Root cause

`Engine.Trigger` (`backend/internal/review/review.go`) gated per-commit idempotency on run *status*, returning any existing non-`Failed` run for `(session, targetSHA)` as "already up to date." That included a row still `ReviewRunRunning` whose reviewer was stopped mid-run without ever calling `Submit` (the only thing that records a verdict / sets `Complete`). Such a row stays `Running` forever, so retrying review for the same commit kept short-circuiting with "Review is already up to date for this commit" until a new commit was pushed.

Per the [maintainer's direction](https://github.com/aoagents/ReverbCode/issues/342#issuecomment-4757428776): an `Alive()`-gated check doesn't cover the reported case, because pane liveness can't distinguish "still actively reviewing" from "pane left open after the user stopped the agent" — both report alive. (`IsAlive` only checks that the zellij session/pane exists, not that the agent process is running.)

## Fix + state-machine hardening

Re-review is **blocked only when a pass actually finished — `Status == Complete`** (set exclusively by `Submit`, after it records a verdict). Every non-`Complete` pass is retryable:

- **`Complete`** → blocked, returned as-is (`Created: false`). Body-independent: an `approved` review may carry no body yet still counts as finished.
- **`Running`, un-verdicted** → never trusted. Superseded (marked `Failed`, body `"superseded by a new review trigger"`), then a fresh pass starts — reusing the live pane via `Notify` if it's still alive, or `Spawn`ing a new one if not.
- **`Failed`** (launch error, or a superseded row) → falls through to a fresh retry.
- **CAS race:** if the supersede `UpdateReviewRunResult` finds the row no longer `Running` (a concurrent `Submit` won), the row is re-read and the now-`Complete` pass returned instead of starting a redundant one.

### Resiliency decisions captured here
- **Block gate = `Status == Complete`** (not verdict-or-body heuristics).
- **Double-click while running** = supersede + re-drive the reviewer on the same pane (no duplicate pane; the per-worker lock + `Notify` guarantee that). Kept as-is and now covered by tests.
- A dedicated `stopped` status (vs. `failed`) for abandoned runs is **deferred** to a follow-up to keep this PR focused on the retry/idempotency state machine.

## Tests (`internal/review`)

- `TestTriggerIsIdempotentForSameCommit`: a `Complete` pass **with an empty body** still blocks (proves the gate is `Status`, not body).
- `TestTriggerReturnsCompletedRunWhenSubmitWinsSupersedeRace` (new): models a concurrent `Submit` landing during supersede; Trigger re-reads and returns the `Complete` pass, no redundant spawn (via a new `fakeStore.beforeUpdate` hook). Previously untested.
- `TestTriggerRespawnsWhenRunningRowHasNoVerdict`: dead pane → stale row `Failed`, fresh `Spawn`.
- `TestTriggerNotifiesLiveReviewerWhenRunningRowHasNoVerdict`: live pane → stale row superseded, pane re-`Notify`d (no second spawn).
- `TestTriggerConcurrentSameWorkerNeverDoubleSpawns`: concurrent triggers never spawn two panes for one commit (#242).

`go build ./...`, `go vet ./internal/review/`, and `go test -race ./internal/review/ ./internal/cli/` all pass. (Pre-existing, unrelated failures in `internal/adapters/runtime/zellij` reproduce on `main` and depend on the local zellij binary.)


backend/internal/review/review.go
```diff
@@ -184,12 +184,32 @@ func (e *Engine) Trigger(ctx context.Context, workerID domain.SessionID) (Trigge
...
+	// liveness can't distinguish "still working" from "pane open, work
+	// stopped." So an un-verdicted Running pass is superseded on retry: it's
+	// marked Failed, and a fresh pass is started below — reusing the pane via
+	// Notify if it's still alive, or spawning a new one if not (#342).
 	if existing, ok, err := e.store.GetReviewRunBySessionAndSHA(ctx, workerID, targetSHA); err != nil {
 		return TriggerResult{}, err
-	} else if ok && existing.Status != domain.ReviewRunFailed {
+	} else if ok && existing.Verdict != domain.VerdictNone {
 		return TriggerResult{Run: existing, ReviewerHandleID: review.ReviewerHandleID, Created: false}, nil
+	} else if ok && existing.Status == domain.ReviewRunRunning {
```
neversettle17-101, line 206: Non-blocking, just flagging the behavior change: previously a repeat `Trigger` against a *genuinely live, in-progress* review (Running, pane alive, not yet submitted) was a pure no-op that returned the existing run. Now it always supersedes — the in-progress run is marked `Failed` ("superseded by a new review trigger") and a fresh run is created, re-`Notify`ing the live pane. Since the only caller is the user-initiated `POST /reviews/trigger` button, this reads as "re-run the review," which is reasonable, and the test comments call it out. Worth confirming nothing auto-polls this endpoint, since on a poll loop it would keep killing an in-flight review before it can submit. From what I can see it's button-only, so this is fine.

neversettle17-101 (commented):
**Approving** (posted as a comment since GitHub blocks self-approval).

Solid fix for #342. The v2 redesign — gating idempotency on a recorded *verdict* rather than run status/pane-liveness — is the right model: pane liveness genuinely cannot distinguish "still working" from "pane open, work abandoned," so trusting only a verdict (which only `Submit` sets) is more robust than the original liveness probe.

What I verified:
- **Concurrent-Submit race is handled correctly.** The supersede uses `UpdateReviewRunResult`, whose SQL is a compare-and-swap (`... WHERE id = ? AND status = 'running'`). If a concurrent `Submit` already transitioned the row to `complete`+verdict, the supersede affects 0 rows, `superseded` is false, and the code re-reads and returns the real verdict instead of starting a redundant pass. Nice.
- **No double-spawn.** `Trigger` holds the per-worker lock across the spawn; losers supersede the prior running row and reuse the live pane via `Notify` (pane considered alive once spawned), so `spawnCount` stays 1. Confirmed by `TestTriggerConcurrentSameWorkerNeverDoubleSpawns`.
- **Unreachable fall-through is actually unreachable.** A `Complete` row always carries a real verdict (`Submit` rejects `VerdictNone` via `verdict.Valid()`), so it always hits the verdict short-circuit and never falls through to a redundant respawn.
- `go build ./...`, `go vet ./internal/review/`, and `go test ./internal/review/` all pass locally.

Test coverage is thorough: live-pane reuse, dead-pane respawn, verdict short-circuit, and the no-double-spawn invariant are all exercised.

One behavior change worth being aware of (non-blocking) noted inline.

backend/internal/review/review.go
```diff
@@ -184,12 +184,32 @@ func (e *Engine) Trigger(ctx context.Context, workerID domain.SessionID) (Trigge
...
+	// Running pass with no verdict may have been interrupted (its execution
+	// stopped, or its pane killed) without ever calling Submit, and pane
+	// liveness can't distinguish "still working" from "pane open, work
+	// stopped." So an un-verdicted Running pass is superseded on retry: it's
+	// marked Failed, and a fresh pass is started below — reusing the pane via
+	// Notify if it's still alive, or spawning a new one if not (#342).
 	if existing, ok, err := e.store.GetReviewRunBySessionAndSHA(ctx, workerID, targetSHA); err != nil {
 		return TriggerResult{}, err
-	} else if ok && existing.Status != domain.ReviewRunFailed {
+	} else if ok && existing.Verdict != domain.VerdictO {
```
neversettle17-101: `domain.VerdictO` is undefined — this must be `domain.VerdictNone`. As written, the package fails to compile (`internal/review/review.go:197:45: undefined: domain.VerdictO`). Appears to be an accidental truncation introduced by the "Test" commit; revert to `existing.Verdict != domain.VerdictNone`.

neversettle17-101 (commented):
**Requesting changes** — the package no longer compiles.

The head commit `ad0e0a4` ("Test") changed `domain.VerdictNone` to `domain.VerdictO` on the verdict short-circuit in `Engine.Trigger`. `domain.VerdictO` does not exist (the defined constants are `VerdictNone`, `VerdictApproved`, `VerdictChangesRequested`), so `go build ./internal/review/` fails:

```
internal/review/review.go:197:45: undefined: domain.VerdictO
```

This looks like an accidental edit — please revert it back to `domain.VerdictNone`. Everything else in the PR (the verdict-gated idempotency redesign from `177eb76`) was correct and previously reviewed as approvable; once this typo is reverted and `go build ./... && go test ./internal/review/` passes again, this is good to go.

neversettle17-101 (commented):
**Approving** (posted as a comment since GitHub blocks self-approval).

The head commit `a0849ec` reverts the broken "Test" commit (`ad0e0a4`), so the tree is now byte-identical to `177eb76` — `git diff 177eb76..a0849ec` is empty. The earlier compile error is gone (`existing.Verdict != domain.VerdictNone` is restored), and `go build ./...` plus `go test ./internal/review/` pass locally.

This is the verdict-gated idempotency redesign I reviewed and found approvable before the typo was introduced:
- idempotency short-circuits only on a recorded verdict (set exclusively by `Submit`);
- an un-verdicted `Running` pass is superseded via a compare-and-swap (`UPDATE ... WHERE status='running'`), which correctly handles the concurrent-`Submit` race (loser re-reads and trusts the verdict);
- the per-worker lock plus pane reuse keeps `spawnCount` at 1 (no double-spawn);
- a `Complete` row always carries a real verdict, so nothing falls through to a redundant respawn.

Tests cover verdict short-circuit, dead-pane respawn, live-pane reuse, and the no-double-spawn invariant. Good to land.

backend/internal/review/review.go
```diff
@@ -184,12 +184,32 @@ func (e *Engine) Trigger(ctx context.Context, workerID domain.SessionID) (Trigge
...
+	// Running pass with no verdict may have been interrupted (its execution
+	// stopped, or its pane killed) without ever calling Submit, and pane
+	// liveness can't distinguish "still working" from "pane open, work
+	// stopped." So an un-verdicted Running pass is superseded on retry: it's
+	// marked Failed, and a fresh pass is started below — reusing the pane via
+	// Notify if it's still alive, or spawning a new one if not (#342).
 	if existing, ok, err := e.store.GetReviewRunBySessionAndSHA(ctx, workerID, targetSHA); err != nil {
 		return TriggerResult{}, err
-	} else if ok && existing.Status != domain.ReviewRunFailed {
+	} else if ok && existing.Verdict != domain.V {
```
neversettle17-101: `domain.VerdictNo` is undefined — must be `domain.VerdictNone`. The package fails to compile (`internal/review/review.go:197:45: undefined: domain.VerdictNo`). Revert to `existing.Verdict != domain.VerdictNone`.

neversettle17-101 (commented):
**Requesting changes** — the package no longer compiles (again).

The head commit `ec0b9b2` ("Test - 2") changed `domain.VerdictNone` to `domain.VerdictNo` on the verdict short-circuit in `Engine.Trigger`. `domain.VerdictNo` does not exist (the defined constants are `VerdictNone`, `VerdictApproved`, `VerdictChangesRequested`), so `go build ./internal/review/` fails:

```
internal/review/review.go:197:45: undefined: domain.VerdictNo
```

This is the same class of breakage as the earlier `ad0e0a4` ("Test") commit that was reverted in `a0849ec` — please revert `ec0b9b2` as well (back to `existing.Verdict != domain.VerdictNone`). The underlying redesign at `a0849ec`/`177eb76` is correct and was approved; once this is reverted and `go build ./... && go test ./internal/review/` pass, it's good to land. Please stop pushing scratch "Test" commits to the PR branch.

neversettle17-101 (commented):
**Approving** (posted as a comment since GitHub blocks self-approval).

This revision recovers from the earlier scratch commits and improves the idempotency model. `go build ./...`, `go vet ./internal/review/`, `go test ./internal/review/`, and `gofmt -l` are all clean.

The re-review block is now gated on `Status == domain.ReviewRunComplete` instead of `Verdict != VerdictNone`. These are equivalent given the existing invariant — `Submit` sets `Complete` and a real verdict together, atomically, via the CAS on `status='running'` — but keying on `Status` reads more clearly ("a commit counts as reviewed once a pass finished") and the state matrix is now spelled out: `Complete` blocks; `Failed` and `Running` are retryable, with a `Running` pass superseded via the CAS.

The standout addition is `TestTriggerReturnsCompletedRunWhenSubmitWinsSupersedeRace`: the new `fakeStore.beforeUpdate` hook flips the row to `Complete` exactly when `Trigger` attempts its supersede `UpdateReviewRunResult`, faithfully modelling a concurrent `Submit` winning the race. It then asserts the supersede no-ops, the now-`Complete` run is returned as-is, and no redundant pass is spawned/notified or inserted. That directly exercises the `!superseded` re-read path that was previously only reasoned about. Nice.

The updated `TestTriggerIsIdempotentForSameCommit` also correctly asserts that an approved `Complete` pass with an empty body still blocks — confirming the gate is on status, not body content.

No issues. Good to land.

Vaibhaav-Tiwari (changes requested):
github currently marks this as not mergeable. It needs a rebase onto main and conflict resolution with the recent review-id changes in backend/internal/review/review.go and review_test.go. after that this should be good to merge.

neversettle17-101:
Thanks for the review. Addressing each point:

**`domain.VerdictO` / `domain.VerdictNo` (compile errors at line 197)** — these were stale typos from the throwaway "Test"/"Update review.go" commits that were pushed onto this branch by another author; they've since been reverted, and the subsequent merge of `main` replaced that gate entirely. The current code at `review.go:204` reads:

```go
} else if ok && existing.Status == domain.ReviewRunComplete {
```

No `VerdictO`/`VerdictNo` references remain anywhere (`grep` clean), and the package compiles — `go build ./...`, `go vet`, and `go test ./...` all pass.

**Supersede-on-retry behavior** — confirmed the concern doesn't apply: nothing auto-polls the trigger path. `Engine.Trigger` has exactly one caller chain — `service/review.Trigger` → the `POST /sessions/{sessionId}/reviews/trigger` HTTP handler (`controllers/reviews.go:73`). No backend loop (lifecycle/observe/reaper) calls it. On the frontend it's a `useMutation` (`SessionInspector.tsx:329`) wired to a button `onClick` (`onTrigger` → line 467), so it fires only on an explicit user click — never on an interval or query refetch. So a retry can't kill an in-flight review out from under a poll loop; "re-run the review" is exactly the intended semantics, and the test comments call it out.

No code change required for these.

neversettle17-101:
Merging this with https://github.com/aoagents/ReverbCode/pull/377

PR #354, opened 2026-06-20
fix: recover terminal reattach after daemon idle

### LINKED ISSUE #350: Terminal can stay stuck reattaching after daemon/session idle
## Problem

The Electron UI can get stuck showing terminal reattach/disconnected state after the app/session has been idle. In the observed case, the daemon was still healthy and ready, but the renderer did not create a fresh terminal mux attach attempt, so the terminal pane stayed blank/reattaching until the app was restarted.

## Evidence

- o status --json reported the daemon as 
eady.
- The daemon process was running from the expected checkout.
- Daemon logs did not show new /mux attach attempts while the UI was stuck.
- A related failure mode is that Electron can also pick up a stale daemon from another checkout via 
unning.json, which leaves the frontend pointed at the wrong process/port.

## Expected behavior

When the daemon is healthy or becomes healthy again, the frontend should recover automatically and reconnect terminal panes without requiring an Electron restart.

## Suggested fix

- Include daemon identity in health/readiness responses.
- Have Electron reject stale/wrong daemons from another checkout.
- Re-probe/re-adopt a compatible existing daemon when renderer asks for daemon status.
- Refresh daemon status while the renderer is not ready, especially after idle/focus restore, so terminal reattach can resume.

Vaibhaav-Tiwari:
## Summary

- add daemon identity to health/readiness probes so Electron can reject stale daemons from another checkout
- have Electron inspect and re-adopt a compatible existing daemon instead of staying on stale status
- refresh renderer daemon status while non-ready and gate terminal attach on daemon readiness

Closes #350

## Tests

- cd backend; go test ./internal/httpd ./internal/cli
- cd frontend; pnpm.cmd run typecheck
- cd frontend; pnpm.cmd run test src/renderer/hooks/useTerminalSession.test.tsx src/renderer/hooks/useDaemonStatus.test.tsx

### DIFF
--- a/backend/internal/cli/status.go
+++ b/backend/internal/cli/status.go
@@ -45,9 +45,11 @@ type daemonStatus struct {
 }
 
 type probeResult struct {
-	Status  string `json:"status"`
-	Service string `json:"service"`
-	PID     int    `json:"pid"`
+	Status           string `json:"status"`
+	Service          string `json:"service"`
+	PID              int    `json:"pid"`
+	ExecutablePath   string `json:"executablePath,omitempty"`
+	WorkingDirectory string `json:"workingDirectory,omitempty"`
 }
 
 func newStatusCommand(ctx *commandContext) *cobra.Command {
--- a/backend/internal/httpd/router.go
+++ b/backend/internal/httpd/router.go
@@ -222,19 +222,26 @@ func localControlRequest(r *http.Request) bool {
 // handleHealthz is the liveness probe: it answers 200 as long as the process is
 // up and serving. It does no dependency checks by design.
 func handleHealthz(w http.ResponseWriter, _ *http.Request) {
-	envelope.WriteJSON(w, http.StatusOK, map[string]any{
-		"status":  "ok",
-		"service": daemonmeta.ServiceName,
-		"pid":     os.Getpid(),
-	})
+	envelope.WriteJSON(w, http.StatusOK, daemonProbePayload("ok"))
 }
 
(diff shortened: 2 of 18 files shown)


areycruzer:
Heads up: I hit a closely related terminal reattach edge on current `main`, but it is restore-specific rather than daemon-idle-specific.

Repro from a fresh temp project/session in Electron:
1. Open a live worker session.
2. Click `Kill session` -> `Confirm kill`.
3. The pane correctly shows `TERMINAL ENDED` with `Restore session`.
4. Click `Restore session`.

Expected: the current session route reattaches to the restored PTY and shows a live terminal again.

Actual: the API restore succeeds and the session is active again (`status: idle`, `isTerminated: false`, same `terminalHandleId`), but the current route stays stuck on `TERMINAL ENDED` with “This terminal process ended, but the session is not marked terminated yet.” The restore button disappears because the session is no longer `terminated`, while the terminal hook still appears to be in `exited`. Navigating away to the board and reopening the session remounts the pane and recovers the terminal.

Likely related code path:
- `TerminalPane` keys `AttachedTerminal` by `terminalHandleId`; restore keeps the same handle, so it does not remount.
- `canRestoreSession` flips false once the restored session status is no longer `terminated`.
- `useTerminalSession` keeps the mux exit state (`exited`) and does not reattach on same `session.id` + same `terminalHandleId` restore.

This feels adjacent to the reattach work in this PR, but distinct: daemon readiness is fine; the missing transition is from exited -> restored/live on the same terminal handle.

PR #373, opened 2026-06-21
fix(desktop): attach to a serving daemon instead of spawning a doomed child

### LINKED ISSUE #367: daemon: Electron-spawned child exits 1 (port already owned) despite pre-check
## Bug

Launching the Electron desktop app while a standalone `ao daemon` already owns port `:3001` logs:

```
daemon already running (pid 5019, port 3001); refusing to start
exit status 1
```

…from the Electron-spawned child daemon, which exits non-zero.

**Source:** Discord `#bug-triaging` thread | **Reported by:** @Tom24 | **Analyzed against:** `26f4d3e` (2026-06-21)
**Confidence:** Medium — the failing code path is confirmed; the exact condition that defeats the Electron pre-check needs one more data point from the reporter (see "Open question").

## Reproduction

1. Start a standalone daemon: `ao daemon` (or `ao start`) → binds `:3001`, writes `~/.ao/running.json`.
2. Launch the Electron desktop app (which spawns its own bundled daemon at boot).
3. Observe: the spawned child logs `daemon already running (pid <N>, port 3001); refusing to start` and exits with status 1.

## Root Cause

> ⚠️ The reporter's initial hypothesis was "Electron's main process **unconditionally** tries to start its own bundled daemon at boot." **This is incorrect.** Electron has a pre-check that should attach to the existing daemon instead of spawning. The defect is the **gap between two checks that read the same handshake file** — Electron's pre-check returns null (so it spawns) while the Go child's check succeeds (so it refuses). The detailed trace follows.

### The Go child's guard (fires — produces the error)

`backend/internal/daemon/daemon.go:46-50`:
```go
if live, err := runfile.CheckStale(cfg.RunFilePath); err != nil {
    return fmt.Errorf("inspect run-file: %w", err)
} else if live != nil && runFileOwnerServing(&http.Client{Timeout: staleProbeTimeout}, config.LoopbackHost, live) {
    return fmt.Errorf("daemon already running (pid %d, port %d); refusing to start", live.PID, live.Port)
}
```
`CheckStale` confirms the run-file's PID is alive; `runFileOwnerServing` verifies it is genuinely serving `/healthz` on the recorded port (so a stale/crashed/reused-PID run-file does not trigger a false refusal). Both passed in the reporter's case → standalone daemon confirmed live → child refuses → exit 1.

### Electron's pre-check (should have caught it — returned null instead)

`frontend/src/main.ts:347` `startDaemonInner()`:
```ts
const launch = resolveDaemonLaunch(...)
const existing = await inspectExistingDaemon(launch)   // line 367
if (existing) {                                         // line 371
    setDaemonStatus(existing)                           // attach, don't spawn
    return daemonStatus
}
// ... spawn() only reached if inspectExistingDaemon returned null (line 394)
```

`inspectExistingDaemon` (`main.ts:259-304`) reads the **same** `~/.ao/running.json`, checks `processAlive(info.pid)`, and probes `/healthz` + `/readyz`. If it finds a live daemon it returns a `ready`/`error` status and Electron **attaches** rather than spawning.

**The bug is: `inspectExistingDaemon` returned `null` for the standalone daemon, even though the Go child's equivalent check succeeded.** Possible conditions that make the pre-check miss while the child's catches:

1. **`runFilePath()` returned null** (`main.ts:261`) — e.g. `AO_RUN_FILE` env mismatch between Electron and the shell that started the standalone daemon, so they read different handshake files. *(Most likely — the two processes may not agree on the run-file path.)*
2. **`readFile` threw** (`main.ts:264-266`) — transient perms / lock / race at the instant Electron booted.
3. **`parseRunFile` returned null** (`daemon-discovery.ts:72`) — malformed JSON or invalid port field in `running.json`.
4. **`processAlive(info.pid)` returned false** (`main.ts:269`) — a platform process-probe quirk (EPERM-vs-ESRCH on signal-based liveness checks) where Node's probe says dead but Go's says alive.
5. **`health.pid !== info.pid`** (`main.ts:272`) — the health probe returned a PID that didn't match the run-file's, so the pre-check bailed to null even though a daemon was genuinely serving.
6. **`daemonIdentityError` path mismatch** (`main.ts:252`) — the standalone daemon's `executablePath` didn't match the bundled binary. *(This would return an `error` state, not null, so it wouldn't trigger a spawn — but worth confirming the reporter didn't see this message instead.)*

### Why this matters

The Electron supervisor and the Go child both treat `~/.ao/running.json` as the single source of truth for "is a daemon already running here," but they reach that truth through **independent parsers and liveness probes**. Any divergence between the two (path resolution, JSON parsing, PID-liveness semantics, health-probe PID matching) produces exactly this symptom: Electron spawns, child refuses, child dies, Electron reports a spawn failure for a daemon that was running fine.

## Open question (needs from reporter)

To pin the exact failure condition, one data point is needed: **what did Electron's `inspectExistingDaemon` actually return, and why?** Specifically:
- Was `~/.ao/running.json` present and readable when Electron booted?
- Did Electron and the standalone daemon agree on the run-file path (`AO_RUN_FILE`)?
- What did the renderer's daemon-status indicator show in the moment before the spawn-attempt error?

Without this, the fix targets a *class* of divergence rather than the specific gap.

## Fix

The defensive fix (covers the whole class): **before spawning, Electron should also try to bind the port** (or probe `127.0.0.1:<port>/healthz` directly, independent of the run-file) — if anything is serving, attach to it rather than spawning. This mirrors what the Go child already does via `runFileOwnerServing` + the `net.Listen` guard in `backend/internal/httpd/server.go:40`:

```go
ln, err := net.Listen("tcp", cfg.Addr())
if err != nil {
    return nil, fmt.Errorf("bind %s (is a daemon already running?): %w", cfg.Addr(), err)
}
```

Electron's `inspectExistingDaemon` trusts the run-file exclusively; adding a port-probe fallback would close the gap for every failure condition above (1–6). Concretely: in `startDaemonInner`, after `inspectExistingDaemon` returns null and before `spawn()`, attempt a direct `fetch` to `http://127.0.0.1:<expectedPort>/healthz`; if it returns a daemon PID, attach instead of spawn.

Secondary: align `runFilePath()` resolution and `parseRunFile` semantics exactly with the Go side (`backend/internal/runfile`), and add a cross-platform `processAlive` parity test.

## Impact

- No data loss, no corruption. The standalone daemon keeps running unharmed.
- Recovery: stop the standalone daemon (`ao stop`) before launching Electron, or kill PID and remove stale `~/.ao/running.json`.
- Only hits the dual-daemon edge case (standalone CLI daemon + desktop app on the same port). No impact on the common path (desktop app launched first).
- UX: Electron surfaces a confusing "daemon failed to start" error for a daemon that was healthy, and never attaches to the running instance.

## Related

- None found. [#324](https://github.com/aoagents/ReverbCode/issues/324) (terminal distortion with Codex/zellij) surfaced in search but is unrelated.

Pritom14:
## Problem

Fixes #367.

Launching the Electron desktop app while a standalone `ao daemon` already owns port `:3001` made the Electron-spawned child daemon log `daemon already running (pid N, port 3001); refusing to start` and exit 1, instead of Electron attaching to the already-running daemon.

The Go child guard and bind guard are correct — they should refuse when a daemon is genuinely serving. The bug is on the Electron side:

- `inspectExistingDaemon` is meant to **attach** instead of spawn, but it trusts `~/.ao/running.json` exclusively. It returns `null` (→ spawn) on any run-file/health divergence: missing/unreadable/unparseable run-file, a recorded PID that isn't alive, or a `/healthz` probe that fails or whose `pid` disagrees with the run-file.
- Between that `null` and `spawn()` there was **no independent port probe** — only an `existsSync(binary)` check. So any divergence made Electron spawn into an already-served port, and the Go child then refused and exited 1.

## Fix

In `startDaemonInner`, **after** the run-file check returns `null` and **before** `spawn()`, probe `http://127.0.0.1:<expectedPort>/healthz` directly (independent of the run-file). If a genuine daemon answers, attach to it — the same `"ready"` `DaemonStatus` shape the run-file path returns — instead of spawning. The expected port is resolved the same way startup does (`AO_PORT` or the default `3001`). Reuses the existing probe helper. Backend (Go) is untouched.

## Testability refactor

The decision logic lived in `main.ts`, which imports `electron` at module load and can't be unit-tested. The attach-or-spawn decision is extracted into a pure, dependency-injected module `shared/daemon-attach.ts` (`expectedDaemonPort`, `parseDaemonProbe`, `resolveDaemonFromRunFile`, `resolveDaemonFromPort`). `main.ts` keeps ownership of fs reads, process signals, `net.fetch`, and the path-identity check, and wires them in. Behavior is preserved.

## Tests — `shared/daemon-attach.test.ts` (27 new)

**Pure unit tests**
- `expectedDaemonPort`: default, valid override, and rejection of out-of-range / non-integer / non-numeric values.
- `parseDaemonProbe`: status↔endpoint match, foreign/missing service, missing/non-integer pid, non-object bodies, non-string identity fields.
- `resolveDaemonFromRunFile`: every branch — no/unparseable run-file, dead PID, health fails, `health.pid` mismatch, "not ready" error, identity error, full attach.
- `resolveDaemonFromPort`: null → spawn, attach on `healthz` ok, probes exactly the expected port.

**Real end-to-end (live loopback HTTP server + real `fetch`)**
- Attaches to a genuinely serving daemon via the direct probe.
- Returns null when the port is closed (→ spawn).
- Does **not** attach to a foreign non-AO server on the port.
- **The #367 scenario**: daemon serving the port but run-file names a **dead PID** → attaches instead of spawning.
- **#367 variant**: run-file PID disagrees with the daemon's reported PID → attaches.
- No regression: run-file fully consistent → still attaches via the run-file path.

## Verification

- `npm run typecheck` (frontend) — clean (strict, no `any`).
- `npm test` (frontend) — **20 files, 205 tests passing** (+27 new).

### DIFF
--- a/frontend/src/main.ts
+++ b/frontend/src/main.ts
@@ -341,6 +341,7 @@ async function startDaemonInner(startEpoch: number): Promise<DaemonStatus> {
 	const directDaemon = await resolveDaemonFromPort({
 		expectedPort: expectedDaemonPort(process.env),
 		probe: readDaemonProbe,
+		identityError: (probe) => daemonIdentityError(launch, probe),
 	});
 	if (startEpoch !== daemonStartEpoch) {
 		return daemonStatus;
--- a/frontend/src/main.ts
+++ b/frontend/src/main.ts
@@ -9,6 +9,13 @@ import { pathToFileURL } from "node:url";
 import { type DaemonLaunchSpec, resolveDaemonLaunch } from "./shared/daemon-launch";
 import { createListenPortScanner, defaultRunFilePath, parseRunFile } from "./shared/daemon-discovery";
 import type { DaemonStatus } from "./shared/daemon-status";
+import {
+	type DaemonProbe,
+	expectedDaemonPort,
+	parseDaemonProbe,
+	resolveDaemonFromPort,
+	resolveDaemonFromRunFile,
+} from "./shared/daemon-attach";
 import { DEFAULT_POSTHOG_HOST, DEFAULT_POSTHOG_PROJECT_KEY } from "./shared/posthog-config";
 import { buildTelemetryBootstrap } from "./shared/telemetry";
 
@@ -157,15 +164,6 @@ const RUN_FILE_POLL_MS = 300;
 // clock reading and ours race within normal scheduling jitter.
 const RUN_FILE_FRESHNESS_SKEW_MS = 2_000;
 const DAEMON_PROBE_TIMEOUT_MS = 2_000;
-const DAEMON_SERVICE_NAME = "agent-orchestrator-daemon";
-
-type DaemonProbe = {
-	status: string;
-	service: string;
-	pid: number;
-	executablePath?: string;
-	workingDirectory?: string;
-};
 
 function runFilePath(): string | null {
 	if (process.env.AO_RUN_FILE) return process.env.AO_RUN_FILE;
@@ -213,17 +211,7 @@ async function readDaemonProbe(port: number, endpoint: "healthz" | "readyz"): Pr
 	try {
 		const response = await net.fetch(`http://127.0.0.1:${port}/${endpoint}`, { signal: controller.signal });
 		if (!response.ok) return null;
-		const body = (await response.json()) as Partial<DaemonProbe>;
-		if (body.status !== (endpoint === "healthz" ? "ok" : "ready")) return null;
-		if (body.service !== DAEMON_SERVICE_NAME) return null;
-		if (typeof body.pid !== "number" || !Number.isInteger(body.pid)) return null;
-		return {
-			status: body.status,
-			service: body.service,
-			pid: body.pid,
-			executablePath: typeof body.executablePath === "string" ? body.executablePath : undefined,
-			workingDirectory: typeof body.workingDirectory === "string" ? body.workingDirectory : undefined,
-		};
+		return parseDaemonProbe(endpoint, await response.json());
 	} catch {
 		return null;
 	} finally {
@@ -258,49 +246,20 @@ function daemonIdentityError(launch: DaemonLaunchSpec, probe: DaemonProbe): stri
 
 async function inspectExistingDaemon(launch: DaemonLaunchSpec): Promise<DaemonStatus | null> {
 	const handshakePath = runFilePath();
-	if (!handshakePath) return null;
-	let contents: string;
-	try {
-		contents = await readFile(handshakePath, "utf8");
-	} catch {
-		return null;
+	let runFileContents: string | null = null;
+	if (handshakePath) {
+		try {
+			runFileContents = await readFile(handshakePath, "utf8");
+		} catch {
+			runFileContents = null;
+		}
 	}
-	const info = parseRunFile(contents);
-	if (!info || !processAlive(info.pid)) return null;
-
-	const health = await readDaemonProbe(info.port, "healthz");
-	if (!health || health.pid !== info.pid) return null;
-	const ready = await readDaemonProbe(info.port, "readyz");
-	if (!ready || ready.pid !== info.pid) {
-		return {
-			state: "error",
-			port: info.port,
-			pid: info.pid,
-			executablePath: health.executablePath,
-			workingDirectory: health.workingDirectory,
-			message: "An AO daemon is already running, but it is not ready yet.",
-		};
-	}
-
-	const identityError = daemonIdentityError(launch, ready);
-	if (identityError) {
-		return {
-			state: "error",
-			port: info.port,
-			pid: info.pid,
-			executablePath: ready.executablePath,
-			workingDirectory: ready.workingDirectory,
-			message: identityError,
-		};
-	}
-
-	return {
-		state: "ready",
-		port: info.port,
-		pid: info.pid,
-		executablePath: ready.executablePath,
-		workingDirectory: ready.workingDirectory,
-	};
+	return resolveDaemonFromRunFile({
+		runFileContents,
+		isProcessAlive: processAlive,
+		probe: readDaemonProbe,
+		identityError: (probe) => daemonIdentityError(launch, probe),
+	});
 }
 
 async function refreshDaemonStatus(): Promise<DaemonStatus> {
@@ -370,6 +329,24 @@ async function startDaemonInner(startEpoch: number): Promise<DaemonStatus> {
 	}
 	if (existing) {
 		setDaemonStatus(existing);
+		return daemonStatus;
+	}
+
+	// Defensive: inspectExistingDaemon only attaches when the run-file agrees with
+	// a live daemon. Any divergence (missing/stale/unparseable run-file, dead PID,
+	// health.pid mismatch) makes it return null — yet a daemon may still be serving
+	// the port. Spawning then would just make the Go child refuse and exit 1. Probe
+	// the expected port directly, independent of the run-file, and attach if a
+	// daemon answers.
+	const directDaemon = await resolveDaemonFromPort({
+		expectedPort: expectedDaemonPort(process.env),
+		probe: readDaemonProbe,
+	});
+	if (startEpoch !== daemonStartEpoch) {
+		return daemonStatus;
+	}
+	if (directDaemon) {
+		setDaemonStatus(directDaemon);
 		return daemonStatus;
 	}
 
--- a/frontend/src/main.ts
+++ b/frontend/src/main.ts
@@ -337,7 +337,9 @@ async function startDaemonInner(startEpoch: number): Promise<DaemonStatus> {
 	// health.pid mismatch) makes it return null — yet a daemon may still be serving
 	// the port. Spawning then would just make the Go child refuse and exit 1. Probe
 	// the expected port directly, independent of the run-file, and attach if a
-	// daemon answers.
+	// daemon answers. The expected port (AO_PORT or the default) is exactly the
+	// port the Go child would bind and collide on — probing a hardcoded 3001 would
+	// miss an AO_PORT override.
 	const directDaemon = await resolveDaemonFromPort({
 		expectedPort: expectedDaemonPort(process.env),
 		probe: readDaemonProbe,
--- a/frontend/src/shared/daemon-attach.ts
+++ b/frontend/src/shared/daemon-attach.ts
@@ -90,13 +90,60 @@ export async function resolveDaemonFromRunFile(deps: RunFileResolveDeps): Promis
 	if (!info || !isProcessAlive(info.pid)) return null;
 
 	const health = await probe(info.port, "healthz");
+	// The recorded PID must match the live daemon; otherwise the run-file points
+	// at the wrong process — return null so the caller falls through to the port
+	// probe rather than trusting a stale handshake.
 	if (!health || health.pid !== info.pid) return null;
-	const ready = await probe(info.port, "readyz");
-	if (!ready || ready.pid !== info.pid) {
+	return readinessStatus(info.port, info.pid, health, probe, identityError);
+}
+
+export type PortProbeResolveDeps = {
+	expectedPort: number;
+	probe: DaemonProber;
+	/** Foreign-daemon check (dev/bundled identity); returns a message, or null when it is ours. */
+	identityError: (probe: DaemonProbe) => string | null;
+};
+
+/**
+ * Attach decision driven by a direct /healthz probe of the expected port,
+ * independent of the run-file (issue #367 backstop). Returns:
+ *   - a "ready" status  → attach to the daemon serving the port,
+ *   - an "error" status → a daemon serves the port but is unusable (not ready, or
+ *                         a foreign binary the identity check refuses); surface it
+ *                         rather than spawn (spawning would only collide and die),
+ *   - null              → nothing genuine answers the port; the caller should spawn.
+ *
+ * This mirrors {@link resolveDaemonFromRunFile}'s post-handshake validation
+ * (/readyz + identity), anchoring on the PID /healthz reports instead of the
+ * run-file's, so attaching via the port is no laxer than attaching via the file.
+ */
+export async function resolveDaemonFromPort(deps: PortProbeResolveDeps): Promise<DaemonStatus | null> {
+	const { expectedPort, probe, identityError } = deps;
+	const health = await probe(expectedPort, "healthz");
+	if (!health) return null;
+	return readinessStatus(expectedPort, health.pid, health, probe, identityError);
+}
+
+/**
+ * Shared tail of both attach paths: given a daemon confirmed serving /healthz on
+ * `port` with PID `pid`, confirm it is ready and is the daemon we expect, and
+ * build the resulting DaemonStatus. Returns an "error" status (never null) — by
+ * here a daemon is definitely occupying the port, so spawning is never the right
+ * move.
+ */
+async function readinessStatus(
+	port: number,
+	pid: number,
+	health: DaemonProbe,
+	probe: DaemonProber,
+	identityError: (probe: DaemonProbe) => string | null,
+): Promise<DaemonStatus> {
+	const ready = await probe(port, "readyz");
+	if (!ready || ready.pid !== pid) {
 		return {
 			state: "error",
-			port: info.port,
-			pid: info.pid,
+			port,
+			pid,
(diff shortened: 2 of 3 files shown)


i-trytoohard:
## Review — fix(desktop): attach to a serving daemon instead of spawning a doomed child

Verified locally: `npm run typecheck` clean, full `npm test` **205/205 across 20 files** (+27 new), no orphan references to the moved symbols. The fix direction is exactly right and matches the #367 recommendation — the testability refactor (extracting pure, DI'd functions out of `main.ts`, which imports `electron` at module load) is the correct call.

The e2e suite is genuinely good — real loopback HTTP server + real `fetch`, covering the exact #367 scenario (stale dead PID → attach), the PID-mismatch variant, foreign-service rejection, and no-regression on the happy run-file path. The `startEpoch` race-guard is correctly placed after the async probe.

One real behavioral inconsistency is worth resolving before merge, plus a minor one.

---

### 🟠 Important — port-probe path skips the foreign-binary identity check

`resolveDaemonFromRunFile` runs `identityError(ready)` and surfaces:

> "Another AO daemon is already running from X; expected Y. Stop the other daemon before using this app."

when the serving daemon's `executablePath` doesn't match the bundled binary (`main.ts:252`). That check exists deliberately to prevent Electron from silently driving a different/older AO build.

`resolveDaemonFromPort` **does not** run that check — it attaches to *any* daemon whose `service` matches, regardless of binary. The probe already has `executablePath`/`workingDirectory` in hand.

Concrete regression path the existing identity guard was built to stop:
1. User runs an old standalone `ao daemon` on `:3001` (different build/commit).
2. Run-file is stale or was never written in a form Electron accepts → `resolveDaemonFromRunFile` returns null.
3. `resolveDaemonFromPort` probes `:3001`, sees a valid health body, and attaches as `"ready"`.
4. Electron now drives the old daemon backend silently — the exact mismatch `daemonIdentityError` was designed to refuse.

**Suggested fix:** give `resolveDaemonFromPort` the same optional `identityError: (probe) => string | null` dependency `resolveDaemonFromRunFile` already accepts, and pass `(probe) => daemonIdentityError(launch, probe)` from `startDaemonInner`. On identity failure, return the same `"error"` DaemonStatus shape (so the user sees the "stop the other daemon" message instead of a silent attach). If skipping the check is intentional (e.g. "the port probe is a last resort; better to attach than spawn-die"), call that out in a comment — but note that attaching to a foreign binary is strictly riskier than surfacing an error, since spawn-die at least *shows* the error.

---

### 🟡 Minor — port-probe path treats `healthz=ok` as `"ready"`, skipping `/readyz`

The run-file path probes both endpoints and, when `readyz` is absent/disagrees, returns an `"error"` state ("An AO daemon is already running, but it is not ready yet.") — it does **not** treat a healthz-only-OK daemon as ready. `resolveDaemonFromPort` probes only `/healthz` and stamps `state: "ready"` immediately.

A daemon mid-startup (alive, not yet ready) would be attached as ready on the port-probe path. Edge case (the port probe is only reached in the unusual run-file-divergent case, and a standalone daemon serving `/healthz` has usually been up a while), but it's a contract divergence worth either closing (also probe `/readyz`, return the "not ready yet" error shape on failure) or documenting.

---

### 🟢 Nits / questions

- **e2e transport:** `realProbe` in the tests uses Node's global `fetch`; `main.ts` uses Electron's `net.fetch`. Acknowledged in the test comments and fine for localhost, but the real transport path (Electron `net` quirks: proxy resolution, cert handling) isn't exercised. Not blocking — the parse logic is what's under test and it's identical.
- **Port selection:** `resolveDaemonFromPort` probes `expectedDaemonPort(process.env)` — the port Electron *would* spawn on. This is the right choice (it's the port the Go child would collide on), but worth a one-line comment confirming that's the intent vs. probing the daemon's *default* 3001 unconditionally.

---

### Summary

Clean, well-tested PR that correctly fixes #367 with the recommended approach. The foreign-binary identity check being silently dropped on the new port-probe path is the one thing I'd want addressed (or explicitly justified) before this lands — it reopens a class of mismatch the existing guard was deliberately built to prevent. The `/readyz` skip is secondary.

(Review only — not an approval.)

Pritom14:
Thanks for the careful review — addressed in `d2ba5fa` (pushed as `8f13889` after the prettier rebase).

### 🟠 Important — foreign-binary identity check on the port-probe path — **fixed**

You're right; silently driving a foreign/older AO build is exactly the regression `daemonIdentityError` exists to prevent. I extracted the shared post-handshake tail (`readinessStatus`) and now run it from **both** paths. `resolveDaemonFromPort` takes the same `identityError` dependency `resolveDaemonFromRunFile` does, and `startDaemonInner` passes `(probe) => daemonIdentityError(launch, probe)`. On identity failure it returns the same `"error"` `DaemonStatus` ("Another AO daemon is already running from X … Stop the other daemon") instead of attaching.

On your "is skipping intentional?" question — no. And as you note, attaching to a foreign binary is strictly worse than spawn-die, but returning the **error** is better than both: spawn-die would just collide on the occupied port and surface a generic "exited with code 1", whereas now the user gets the actionable identity message.

### 🟡 Minor — `/readyz` skip — **fixed**

Same refactor closes this. The port path now probes `/readyz`, checks `ready.pid === health.pid`, and returns the "An AO daemon is already running, but it is not ready yet." error shape when readiness fails — identical to the run-file path. A mid-startup daemon is no longer stamped `"ready"`.

The port path anchors on the PID `/healthz` reports (there's no run-file PID to anchor on in this path), so attaching via the port is no laxer than attaching via the file.

### 🟢 Nits

- **Port selection**: added a clarifying comment — we probe `expectedDaemonPort(process.env)` precisely because that's the port the Go child *would* collide on; probing a hardcoded 3001 would miss an `AO_PORT` override.
- **e2e transport** (global `fetch` vs electron `net.fetch`): left as-is per your "not blocking" — the parse/decision logic is what's under test and is shared verbatim. Noted in the test comments.

### New tests

`daemon-attach.test.ts` now 30 tests: added port-path **not-ready → error**, port-path **foreign-binary identity → error** (unit), and a **real loopback-server foreign-AO-binary** scenario (e2e) asserting an error surfaces rather than a silent attach. `npm run typecheck` clean, `npm test` 208/208.

frontend/src/main.ts
```diff
@@ -373,6 +332,27 @@ async function startDaemonInner(startEpoch: number): Promise<DaemonStatus> {
...
 
+	// Defensive: inspectExistingDaemon only attaches when the run-file agrees with
+	// a live daemon. Any divergence (missing/stale/unparseable run-file, dead PID,
+	// health.pid mismatch) makes it return null — yet a daemon may still be serving
+	// the port. Spawning then would just make the Go child refuse and exit 1. Probe
+	// the expected port directly, independent of the run-file, and attach if a
+	// daemon answers. The expected port (AO_PORT or the default) is exactly the
+	// port the Go child would bind and collide on — probing a hardcoded 3001 would
+	// miss an AO_PORT override.
+	const directDaemon = await resolveDaemonFromPort({
```
Pritom14, line 343: **Blocking:** this port-probe backstop is added here in `startDaemonInner`, but the symmetric refresh path (`refreshDaemonStatus` → `inspectExistingDaemon`, line ~277) was *not* given it — it still calls only `resolveDaemonFromRunFile`.

Result: after this code attaches via the port probe in a divergence case (dead PID / `health.pid` mismatch / missing run-file), the renderer's next `daemon:getStatus` poll (≤10s, `useDaemonStatus.ts`) runs `refreshDaemonStatus`, the run-file is still diverged → `null`, and since the state was `ready` it flips to `stopped`/"AO daemon is no longer reachable" and `applyDaemonStatus` does `setApiBaseUrl(null)`, dropping REST to a daemon that's actually serving.

The attach is undone within seconds for exactly the cases this probe was added to fix. Please share one `resolveDaemon({run-file → port})` helper between `startDaemonInner` and `inspectExistingDaemon`/`refreshDaemonStatus`.

frontend/src/shared/daemon-attach.ts
```diff
@@ -0,0 +1,172 @@
...
+export type DaemonProber = (port: number, endpoint: "healthz" | "readyz") => Promise<DaemonProbe | null>;
+
+/**
+ * The port a freshly spawned daemon is expected to bind: AO_PORT when set and
+ * valid, otherwise the daemon's default. Used to probe for an already-serving
+ * daemon before spawning a child that would only refuse and exit.
+ */
+export function expectedDaemonPort(env: Record<string, string | undefined>): number {
+	const configured = env.AO_PORT ? Number(env.AO_PORT) : NaN;
+	return Number.isInteger(configured) && configured >= 1 && configured <= 65535 ? configured : DEFAULT_DAEMON_PORT;
```
Pritom14, line 46: Minor: `AO_PORT=0` falls back to `3001` here, but `0` requests an OS-assigned ephemeral port (see `daemon-discovery.ts:1-3`), so a spawned child wouldn't collide on `3001`. Probing `3001` in that case could attach to an unrelated default-port daemon rather than spawning the intended ephemeral one. Niche, but a comment documenting the intended behavior would help.

Pritom14 (commented):
**This review requests changes** (posted as a COMMENT because GitHub does not let the PR author submit a formal Request-Changes/Approve).

Solid diagnosis and a clean pure/DI refactor with thorough tests. One blocking issue: the port-probe backstop was added to `startDaemonInner` but **not** to the refresh path, so it gets undone on the next status poll.

### Blocking: `refreshDaemonStatus` reverts a port-probe attach to "stopped"

`startDaemonInner` now consults run-file *then* the direct port probe. But `refreshDaemonStatus` → `inspectExistingDaemon` still consults **only** `resolveDaemonFromRunFile`. In the #367 divergence cases (run-file names a dead PID, or `health.pid` mismatch, or missing run-file) while a daemon genuinely serves the port:

1. `startDaemonInner` attaches via `resolveDaemonFromPort` → `ready`, REST pointed at the port.
2. Renderer polls `daemon:getStatus` every 10s while ready (`useDaemonStatus.ts`).
3. `refreshDaemonStatus` runs the run-file path only → `null` (still diverged).
4. Since prior state was `ready`, it flips to `{state:"stopped", message:"AO daemon is no longer reachable."}` and pushes it.
5. `applyDaemonStatus` then `setApiBaseUrl(null)` — tears down REST to a reachable daemon.

So within ≤10s the fix is undone, for exactly the divergent run-file states the port probe was added to rescue. Pre-PR, `start` and `refresh` were symmetric (run-file only); this PR makes them asymmetric. Suggest a single `resolveDaemon({run-file, then port})` helper shared by both paths so they can't drift.

### Minor

- `expectedDaemonPort` maps `AO_PORT=0` to `3001`, but `AO_PORT=0` asks for an ephemeral port (per `daemon-discovery.ts`), so a spawned child wouldn't collide on 3001 — probing 3001 could attach to an unrelated daemon. Niche; worth a note.

Full review in review.md. Nice work on the e2e loopback tests and anchoring `/readyz` on `health.pid`.

i-trytoohard:
## Re-review — both concerns fully resolved ✅

Verified locally against merged main: `npm run typecheck` clean, **220/220 tests across 22 files** (was 205 — the new identity/readiness tests are in).

### 🟠 Identity check on port-probe path → FIXED

`resolveDaemonFromPort` now accepts `identityError: (probe) => string | null` and `main.ts` wires `(probe) => daemonIdentityError(launch, probe)` in. The exact regression path I flagged — silently attaching to a foreign binary — now surfaces the "Another AO daemon is already running from X" error instead. New test: `"reports an identity error (not a silent attach) for a foreign daemon binary on the port"`.

### 🟡 /readyz skip → FIXED

Both attach paths now converge into a shared `readinessStatus()` tail function that probes `/readyz`, checks `ready.pid === pid`, and runs `identityError`. The port-probe path is no longer laxer than the run-file path — structurally identical post-handshake validation. New test: `"reports an error (not a spawn) when the serving daemon is not ready yet"`.

### 🟢 Port selection docs → ADDRESSED

The comment now explains why the expected port (not a hardcoded 3001) is probed.

### Bonus: clean DRY refactor

Extracting `readinessStatus()` as the shared tail of both `resolveDaemonFromRunFile` and `resolveDaemonFromPort` eliminates the duplicated validation logic from the original. No more "two paths with different strictness" — one function, one contract.

**Verdict: ship it.** Both concerns closed, comprehensive test coverage for both new checks, clean refactor. No further changes needed.

(Review only — not an approval.)

Issue #366, opened 2026-06-21
frontend: brand text overlapped by TitlebarNav toggle on board routes


i-trytoohard:
## Bug

On the board routes (`/` and `/projects/:id`), the sidebar's brand text ("Agent Orchestrator") is overlapped by the fixed `TitlebarNav` toggle cluster — the panel-toggle icon sits directly on top of the brand text.

**Source:** Discord `#bug-triaging` thread | **Reported by:** @Tom24 | **Analyzed against:** `26f4d3e` (2026-06-21)
**Confidence:** High — every cited line traced and verified against current code.

## Reproduction

1. Launch the desktop app (or run the renderer).
2. Navigate to `/` (the default board) or any `/projects/<id>` route — i.e. any route where the `ShellTopbar` is suppressed.
3. Observe: the `TitlebarNav` cluster (toggle + history buttons) overlaps the "Agent Orchestrator" brand text in the sidebar header.

## Root Cause

Three independent layout decisions collide **only** on board routes:

**1. `TitlebarNav` is a fixed overlay, not in flow.**
`styles.css:534`:
```css
.titlebar-nav {
    position: fixed;
    top: 0;
    left: 79px;
    z-index: 20;
    height: 56px;
    ...
}
```
It is rendered **once** by the shell on every route (`_shell.tsx:189`), by design, so its buttons never move when the sidebar collapses. The cluster claims the region `0–56px` tall, starting at `left:79px`.

**2. On routes with a topbar, two insets keep the fixed overlay in its own lane.**
- `styles.css:553`: `.dashboard-app-header.is-under-titlebar-nav { padding-left: 180px; }` — pushes the header content past the cluster.
- `Sidebar.tsx:165`: `className={cn(... underTopbar ? "top-14 h-[calc(100svh-3.5rem)]!" : "top-0 h-svh!")}` — the sidebar drops below the 56px topbar strip.

With both insets active, the fixed cluster lives in the `56px` strip **above** the sidebar — no overlap.

**3. On board routes, both insets are switched off in one place.**
`_shell.tsx:49`:
```ts
const isBoardRoute = pathname === "/" || /^\/projects\/[^/]+$/.test(pathname);
```
- `_shell.tsx:158`: `{!isBoardRoute && <ShellTopbar />}` — no topbar rendered on board routes.
- `_shell.tsx:170`: `underTopbar={!isBoardRoute}` — so the sidebar takes the `top-0` branch (`Sidebar.tsx:165`).

[...]

PR #263, opened 2026-06-16
feat: surface SCM summaries in desktop

whoisasx:
## Problem

The backend SCM observer already persists PR, CI, review, and mergeability facts, and `/api/v1/sessions` derives session status from those facts. The desktop app was not reliably surfacing that state: it treated `/events` as the main live signal, gated PR detail fetches on mock-style `session.pullRequest` data, and masked some backend statuses such as `merged`.

## Scoped V1 behavior

- Desktop fetches current sessions from `GET /api/v1/sessions` and treats backend `session.status` as authoritative.
- Desktop keeps `GET /api/v1/events` open as an invalidation stream only.
- Visible/focused desktop surfaces fetch concise SCM summaries from `GET /api/v1/sessions/{sessionId}/pr`.
- The summary API exposes user-facing SCM details only:
  - PR identity, title, branches, author, repo, provider links, and observed timestamps.
  - CI aggregate state plus failing/cancelled check names and links.
  - Review decision plus reviewer IDs, counts, and links for unresolved human comments.
  - Mergeability state, concise reason codes, and PR link.
- The desktop understands `no_signal`, preserves backend `merged` even for terminated rows, and uses a neutral `unknown` fallback for unrecognized status values.

## Backend changes

- Added a session-service PR summary read model assembled from existing persisted PR, check, review-thread, and comment facts.
- Expanded `GET /api/v1/sessions/{sessionId}/pr` to return `SessionPRSummary` items while leaving the PR claim response on the existing aggregate `SessionPRFacts` shape.
- Added enum metadata for session status, PR state, CI state, failing-check status, review decision, and mergeability in the code-first OpenAPI contract.
- Regenerated `backend/internal/httpd/apispec/openapi.yaml` and `frontend/src/api/schema.ts`.
- Added tests covering the summary shape, failing check links without log tails, unresolved human reviewers without comment bodies, and conflict state without invented conflict files.

## Frontend changes

- Added `useSessionScmSummary(sessionId)` over the generated `SessionPRSummary` type.
- Invalidates both workspace/session queries and SCM-summary queries from daemon status changes, SSE events, SSE reconnects, and API base URL changes.
- Session inspector now fetches PR summaries without requiring pre-existing `session.pullRequest` data and renders PR, CI, review, and mergeability sections.
- Board cards fetch compact visible-session SCM context and show PR/CI/review/mergeability highlights.
- Pull requests page is backed by real session-scoped PR summaries rather than `WorkspaceSession.pullRequest`.
[...]

PR #374, opened 2026-06-21
test(e2e): guard sidebar brand vs macOS titlebar cluster (#366)

Pritom14:
## Context

While rebasing onto `main`, I found that **#366 is already fixed upstream**. #263 ("surface SCM summaries in desktop") changed the shell to render `ShellTopbar` on *every* route and removed the old `isBoardRoute` branch. The sidebar is therefore always `top-14` (below the 56px macOS titlebar band), so the brand never lands in the fixed `TitlebarNav` cluster's lane — the original overlap is structurally gone.

Verified live on current `main` (forced macOS UA): brand at y71, cluster y0–56 → **no overlap, wordmark not truncated**.

My earlier code fix (board-route brand offset) is now obsolete — worse, it would *break* the layout on main (the always-on topbar already drops the sidebar 56px; adding the offset again would double-count). So I dropped it.

## What this PR now is

A focused **e2e regression guard** that locks the invariant in so the bug can't silently return (e.g. if a topbar-less route is reintroduced):

`frontend/e2e/titlebar-brand.spec.ts` — drives the live renderer under a macOS UA and asserts, on the routes #366 named:
- **home board route (`/`)** — brand does not overlap the fixed cluster **and** the "Agent Orchestrator" wordmark is not truncated (`scrollWidth <= clientWidth`).
- **project board route (`/projects/:id`)** — same invariant after in-app navigation.
- **board → session transition** — the persistent brand does not jump position and stays readable.

## Verification

- All 3 tests pass against current `main` with **no source changes**.
- `npm run typecheck` clean.
- No conflicts (branch is `main` + one test file).

### DIFF
--- a/frontend/e2e/titlebar-brand.spec.ts
+++ b/frontend/e2e/titlebar-brand.spec.ts
@@ -1,0 +1,82 @@
+import { expect, test, type Locator, type Page } from "@playwright/test";
+
+// Regression guard for #366 (macOS): the sidebar's "Agent Orchestrator" brand
+// must never sit under the fixed TitlebarNav cluster, and the wordmark must stay
+// readable. The original bug was board routes (`/` and `/projects/:id`) having no
+// topbar, so the sidebar stayed at top-0 and the brand landed in the cluster's
+// 56px lane. It is now fixed structurally — the shell renders the topbar on every
+// route, so the sidebar always hangs below the titlebar band — and these tests
+// lock that invariant in: if a topbar-less route is ever reintroduced, they fail.
+//
+// macOS-only: TitlebarNav (and the bug) gate on navigator.userAgent looking like
+// a Mac, read once at module load. Force a Mac UA so this is deterministic
+// regardless of the host/CI OS.
+test.use({
+	userAgent:
+		"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36",
+});
+
+const brand = (page: Page) => page.getByText("Agent Orchestrator", { exact: true });
+
+// Two boxes overlap iff they intersect on both axes.
+function overlaps(a: { x: number; y: number; width: number; height: number }, b: typeof a) {
+	return a.x < b.x + b.width && a.x + a.width > b.x && a.y < b.y + b.height && a.y + a.height > b.y;
+}
+
+// The brand <span> has `truncate` (overflow:hidden), so it stays "visible" even
+// when clipped to nothing. Compare scroll vs client width to prove the wordmark
+// is actually fully rendered, not just present-but-clipped.
+async function isTruncated(span: Locator) {
+	return span.evaluate((el) => el.scrollWidth > el.clientWidth + 1);
+}
+
+async function expectBrandClearsCluster(page: Page) {
+	const cluster = page.locator(".titlebar-nav");
+	await expect(cluster).toBeVisible();
+	const span = brand(page);
+	await expect(span).toBeVisible();
+
+	const clusterBox = await cluster.boundingBox();
+	const brandBox = await span.boundingBox();
+	expect(clusterBox).not.toBeNull();
+	expect(brandBox).not.toBeNull();
+
+	expect(overlaps(brandBox!, clusterBox!)).toBe(false);
+	expect(await isTruncated(span)).toBe(false);
+}
+
+test("home board route: brand clears the macOS titlebar cluster and stays readable", async ({ page }) => {
+	await page.goto("/");
+	await expect(page.getByText("Projects")).toBeVisible();
+	await expectBrandClearsCluster(page);
+});
+
+test("project board route: brand clears the macOS titlebar cluster and stays readable", async ({ page }) => {
+	await page.goto("/");
+	await expect(page.getByText("Projects")).toBeVisible();
+
+	// In-app nav to /projects/:id (a hard load boots the router at the board).
+	await page.getByRole("button", { name: "Open api-gateway dashboard" }).click();
+	// The active project row marks itself aria-current=page once navigation lands.
+	await expect(page.locator('[aria-current="page"]')).toBeVisible();
+
+	await expectBrandClearsCluster(page);
+});
+
+test("brand stays put and readable when navigating board → session", async ({ page }) => {
+	await page.goto("/");
+	await expect(page.getByText("Projects")).toBeVisible();
+
+	const boardBrandBox = await brand(page).boundingBox();
+	expect(boardBrandBox).not.toBeNull();
+
+	await page.getByRole("button", { name: "Open Split terminal mux responsibilities" }).click();
+	await expect(page.locator(".dashboard-app-header")).toBeVisible();
+
+	const sessionBrandBox = await brand(page).boundingBox();
+	expect(sessionBrandBox).not.toBeNull();
+	// Persistent shell element: no vertical/horizontal jump across the transition.
+	expect(Math.abs(sessionBrandBox!.x - boardBrandBox!.x)).toBeLessThanOrEqual(1);
+	expect(Math.abs(sessionBrandBox!.y - boardBrandBox!.y)).toBeLessThanOrEqual(1);
+	await expectBrandClearsCluster(page);
+});


i-trytoohard:
## Review — fix(sidebar): clear macOS titlebar cluster from board-route brand (#366)

Verified locally: `npm run typecheck` clean, unit suite **178/178 across 19 files**. Couldn't run the e2e suite in this environment (no display server), but the regression proof documented in the PR body (neutralize fix → brand `x=50` fail; restore → `x=220` pass) is the right methodology and the assertions are correct.

Mechanism is sound, scoping is tight (`isBoardRoute && isMac` only), specificity claim holds (`group-data-[collapsible=icon]:px-1.5` is a 3-class/attribute selector that cleanly overrides `pl-[180px]` in the icon rail). The `isMac` gate matches `TitlebarNav.tsx:45` (`if (!isMac) return null`) — so the inset only applies where the cluster actually renders. Good.

One real UX concern, plus a minor test gap.

---

### 🟠 Important — the fix trades "overlapped" for "invisible" at default sidebar width

At the default 240px sidebar width, a 180px left padding on `SidebarHeader` leaves ~53px of content area. After the inner div's `px-2` (16px) and the logo + gap (32px), there are **~5px left for the "Agent Orchestrator" text** — which has `truncate`, so it clips to effectively nothing. The brand becomes logo-only.

The bug was "brand text unreadable behind toggle icon." This fix makes the brand text **unreadable behind a truncation clip**. We solved the overlap but the user still can't read the brand text at the default width on the default route. The PR honestly flags this, but it's worth deciding explicitly whether logo-only is acceptable on the landing page.

**Why 180px is more than needed:** The cluster spans `left:79px` to right edge `167px` (79 + 3×28 + 2×2). The logo sits at `x≈8–30px` — it **never overlaps the cluster**. Only the text (natural start `x≈50px`) collides with the cluster starting at 79px. So the text only needs to shift ~117px to clear 167px, not 180px. The 180px value (mirroring the header's `is-under-titlebar-nav`) over-clears by ~63px and pushes the logo unnecessarily.

**Possible refinements (any one):**
- **(a)** Apply the inset to just the text `<span>`, not the whole `SidebarHeader`. Leave the logo at `x≈8`, push only the text past 167px (`pl-[130px]` or `ml-[120px]` on the span). Logo stays, text is readable, no wasted space.
- **(b)** Board-route minimum sidebar width — bump `--ao-sidebar-w` default or `min` (currently 200) on board routes so the 180px inset doesn't eat the brand. Heavier change.
- **(c)** Accept logo-only on board routes as intentional (the board already has prominent branding elsewhere). If so, drop the `<span>` entirely on board routes to avoid rendering a 5px clipped fragment.

If the maintainer is fine with logo-only at default width, this PR is correct as-is — just confirm the decision.

---

### 🟡 Minor — e2e `toBeVisible()` doesn't catch the truncation

The test asserts the brand element is `toBeVisible()` and positioned past the cluster. But `truncate` (`overflow: hidden; text-overflow: ellipsis`) keeps the element "visible" even when the text is clipped to zero characters. So the regression test passes whether the brand reads "Agent Orchestrator" or shows nothing. If the truncation tradeoff matters (see above), consider asserting the text element's `scrollWidth <= clientWidth` (no overflow = not truncated) at the default sidebar width, so the test guards against future regressions into "text present but clipped."

---

### 🟢 Nits

- **E2E selector brittleness:** `"Open Split terminal mux responsibilities"` as the non-board route navigation target is fragile (depends on mock-data content). Pre-existing problem (the PR notes baseline e2e failures from stale selectors), not introduced here — but worth a TODO to replace with a stable `data-testid`.
- **Comment accuracy:** `Sidebar.tsx:170` says "left 79 → right edge 167" — correct, but the code applies 180px (not 167px). Minor, but someone reading later will wonder why 180 not 167. One line explaining "180 mirrors the topbar inset" would help.

---

### Summary

Clean, minimal, correctly-scoped fix that resolves the overlap. The open question is whether the resulting truncation at default width is acceptable or should be refined (option (a) — inset only the text — would be my pick: smallest change, both logo and text survive). Not a correctness bug, but a UX decision the maintainer should make consciously before merge.

(Review only — not an approval.)

Pritom14:
Thanks @i-trytoohard — this was the right catch. Rather than refine the 180px inset, I changed the approach based on your geometry point (the logo never overlaps the cluster — only the text does). Pushed in `1bd38a7`.

**New approach: drop the brand *below* the 56px titlebar band instead of shoving it right.**

On board routes the brand row now gets `pt-[70px]` (macOS only) so it sits at the *same y* it occupies on routes that have a topbar (`top-14` + `pt-3.5` = 70px), with the logo + text at their natural left position. Net result:

- **Full "Agent Orchestrator" wordmark stays readable** at the default 240px width — no truncation.
- Logo and text stay together at the left (no cramming).
- The top strip reads as a proper macOS titlebar (traffic lights + the fixed cluster live there; brand sits below).
- **Bonus:** the brand no longer jumps vertically when navigating board↔session — it was previously at y15 on board and y70 elsewhere; now it's y70 on both.

I compared the candidates by rendering each against the live board (180px / text-only inset / push-down / logo-only). Push-down was the only one preserving the full wordmark, so I went with it.

Addressing your other points:

- **🟡 Test gap — fixed.** The test now asserts (a) the brand does **not** overlap the cluster (axis-agnostic rect check), (b) the wordmark is **not truncated** (`scrollWidth <= clientWidth`), and (c) it does **not jump** position navigating board→session. I verified all three fail when the offset is neutralized, so they're genuine guards (the truncation assertion specifically guards against the "present but clipped" regression you flagged).
- **🟢 Comment accuracy — fixed.** The `SidebarHeader` comment now explains the 70px = 56px titlebar band + 14px breathing, matching the topbar-route position.
- **🟢 Selector brittleness — acknowledged, left as-is.** `"Open Split terminal mux responsibilities"` depends on mock-data content; it's the same brittleness as the existing (currently-stale) e2e selectors. Worth a follow-up to add stable `data-testid`s across the e2e suite, but out of scope for this fix.

Icon rail is unaffected (the cluster sits beyond its 48px width); the offset resets to `pt-3.5` there via the higher-specificity `group-data-[collapsible=icon]` variant — verified via computed style (expanded `pt:70px`, collapsed `pt:14px`).

Verification: `npm run typecheck` clean, unit suite 178/178, both e2e tests pass.

Pritom14:
Heads-up @i-trytoohard — the PR changed substantially after rebasing onto `main`.

I discovered **#366 is already fixed upstream** by #263: the shell now renders the topbar on every route (the old `isBoardRoute` branch is gone), so the sidebar always hangs below the 56px titlebar band and the brand never overlaps the cluster. Confirmed live on main (brand y71, cluster y0–56, no overlap, not truncated).

That made my code fix obsolete — and it would actually have broken the layout on main (the always-on topbar already drops the sidebar 56px, so re-applying my offset would double-count). So I dropped both code commits and reduced this PR to just the **e2e regression guard**, rebased clean on main (1 commit, no conflicts).

The test now covers both board routes the issue named (`/` and `/projects/:id`) plus a board→session transition, asserting no overlap with the fixed cluster and no truncation of the wordmark. It passes against current main with zero source changes, so it documents and locks in the now-structural fix. Your earlier truncation/test-gap points are addressed by the not-truncated assertions.

i-trytoohard:
## Re-review — completely different approach, and it's the right one ✅

The original PR (option B: 180px padding hack on `SidebarHeader`) is gone. Instead the fix went **structural**: `isBoardRoute` is removed entirely from `_shell.tsx`, `<ShellTopbar />` now renders unconditionally on every route, and `underTopbar` is hardcoded `true`. The sidebar always hangs below the 56px titlebar strip, so the brand row never enters the cluster's lane. Root cause eliminated, no symptom patch.

This resolves the 🟠 truncation tradeoff I flagged — there's no inset to truncate with. The brand text renders at its natural position below the titlebar band, fully readable at any sidebar width. Better than any of the three options I suggested in the issue.

The structural fix is already merged to main. This PR now carries only the regression test, which is well-built:
- `overlaps()` — proper 2D bounding-box collision (not just x-coordinate). ✅
- `isTruncated()` — `scrollWidth > clientWidth + 1` check. ✅ Directly addresses my 🟡 nit that `toBeVisible()` doesn't catch clipped text.
- Tests both board routes (`/` and `/projects/:id`) + a navigation-persistence guard (brand doesn't jump across route transitions).
- Test comment clearly documents the invariant being locked in.

**Verdict: ship it.** The structural fix is superior to the padding hack, and the test guards both the overlap and the truncation regression classes.

(Review only — not an approval.)

PR #377, opened 2026-06-21
Refactoring Reviews feature code and idempotency check

neversettle17-101:
## Summary
- Supersedes #340 by moving AO-internal changes_requested delivery out of review.Engine and into ReviewService + lifecycle.Manager.ApplyReviewResult.
- ReviewService now persists the review_run result first, invokes the lifecycle reducer synchronously, then stamps delivered_at/status=delivered only after lifecycle reports delivery.
- Lifecycle labels review feedback origins as [AO reviewer] and [PR reviewer @user], and AO review nudges use sendOnce with per-pass keys/signatures so restart-safe dedup remains the single primitive.
- Adds review_run.delivered_at migration, sqlc/API regeneration, and focused tests for reducer relevance/dedup plus service persist-then-call-then-stamp sequencing.

## Go rewrite boundary
- Legacy behavior ported/preserved: review feedback nudges are lifecycle reactions, matching the PR observation reaction path in backend/internal/lifecycle/reactions.go rather than direct service-to-runtime sends.
- Go service ownership: ReviewService owns review_run writes; lifecycle owns in-process side effects only; controller stays transport-only.
- Invariants preserved: CLI remains a daemon HTTP client; derived status is still stored only as durable facts; worker nudges use lifecycle sendOnce persisted PR signatures and do not read change_log.

## Verification
- gh pr diff 340
- npm run sqlc
- env GOCACHE=/private/tmp/ao-go-cache npm run api
- env GOCACHE=/private/tmp/ao-go-cache go build ./...
- env GOCACHE=/private/tmp/ao-go-cache go vet ./...
- env GOCACHE=/private/tmp/ao-go-cache go test ./... (passed with escalation; sandboxed run failed only on expected port/user-config/zellij permissions)


backend/internal/review/review.go
```diff
@@ -276,17 +272,9 @@ func (e *Engine) Trigger(ctx context.Context, workerID domain.SessionID) (Trigge
...
-// On a changes_requested verdict, Submit also messages the worker session with
-// the review feedback directly, so the worker learns about it event-driven
-// rather than via the SCM poll loop (which never observes CHANGES_REQUESTED for
-// self-reviews or COMMENT-state reviews; issue #337). When a GitHub review id is
-// known, it is included so the worker knows exactly which review to address and
-// reply to.
+// Submit records the reviewer's result for a specific worker review pass. The
+// API service owns any post-persist lifecycle delivery; the core engine only
+// validates ownership and persists the result.
 func (e *Engine) Submit(ctx context.Context, workerID domain.SessionID, runID string, verdict domain.ReviewVerdict, body, githubReviewID string) (domain.ReviewRun, error) {
```
neversettle17-101: `Engine.Submit` is no longer reached from production: the controller goes through `service.Submit`, which reimplements validation + the persist state machine directly against the store. The two have diverged (this one rejects anything but `running`; the service handles `complete`/`delivered` idempotently), and this method is now only covered by engine tests. Consider deleting it or having the service delegate here, so the submit logic doesn't drift in two places.

backend/internal/lifecycle/reactions.go
```diff
@@ -103,11 +138,49 @@ func (m *Manager) ApplyPRObservation(ctx context.Context, id domain.SessionID, o
...
+	if r.Body != "" {
+		msg += "\n\n" + domain.SanitizeControlChars(r.Body)
+	}
+	key := "review:" + r.PRURL + ":ao:" + r.RunID
+	sig := strings.Join([]string{r.TargetSHA, r.RunID, r.GithubReviewID, r.Body}, "\x00")
+	sent, err := m.sendOnce(ctx, workerID, r.PRURL, key, sig, msg, reviewMaxNudge)
+	if err != nil {
+		return ReviewDeliveryNoop, err
+	}
+	if !sent {
```
neversettle17-101: Dead branch: `if !sent { return ReviewDeliverySent, nil }` and the following `return ReviewDeliverySent, nil` return the same value. Returning `Sent` on dedup is the intended behavior (so `delivered_at` gets stamped), so this `if` block can just be dropped.

backend/internal/service/review/review.go
```diff
@@ -46,7 +82,89 @@ func (s *Service) Trigger(ctx context.Context, workerID domain.SessionID) (revie
...
+			return domain.ReviewRun{}, err
+		}
+		if !updated {
+			return domain.ReviewRun{}, fmt.Errorf("%w: review run %q is not running", ErrInvalid, runID)
+		}
+		run.Status = domain.ReviewRunComplete
+		run.Verdict = verdict
+		run.Body = body
+		run.GithubReviewID = githubReviewID
+	case domain.ReviewRunComplete:
```
neversettle17-101, line 124: In the `complete` branch you update `run.Body`/`run.GithubReviewID` in memory from the new args but never persist them (`UpdateReviewRunResult` only fires on status `running`). On a retry that supplies a fuller body, lifecycle delivers the new text while the stored row keeps the old. Minor, but worth either persisting or ignoring the new args on re-submit for consistency.

neversettle17-101 (commented):
**Verdict: Approve (ready) — with non-blocking suggestions.**

This is a clean refactor. Moving changes_requested delivery out of `review.Engine` and into `ReviewService` + `lifecycle.Manager.ApplyReviewResult` is the right boundary: the service now persists the result first, runs the lifecycle reducer synchronously, then stamps `delivered_at`/`status=delivered` only after lifecycle confirms delivery. The persist-then-call-then-stamp sequencing and the restart-safe dedup (per-pass `sendOnce` key/signature + `delivered_at` guard) are well thought through and well tested. Migration `0018` correctly slots after main's `0017`, and the sqlc/openapi/schema regenerations are consistent. Build, vet, and the focused tests pass locally.

No correctness blockers. A few maintainability nits below.

### Suggestions (non-blocking)
1. **`review.Engine.Submit` is now orphaned in production.** The controller calls `service.Submit`, which fully reimplements validation + the persist state machine against the store directly and no longer delegates to the engine. The two have already diverged (engine only accepts `running`; service handles running/complete/delivered idempotently). Engine.Submit is now exercised only by its own tests. Consider removing it (and its tests) or having the service reuse it, to prevent future drift.
2. **Dead conditional in `ApplyReviewResult`** — both branches return the same value.
3. **Complete-branch updates in-memory body/reviewID without persisting** — minor; flagged inline.

backend/internal/review/review.go
```diff
@@ -188,12 +186,29 @@ func (e *Engine) Trigger(ctx context.Context, workerID domain.SessionID) (Trigge
 		return TriggerResult{}, err
 	}
 
-	// Idempotency: return a non-failed pass as-is. Failed passes stay visible
-	// but can be retried after the user fixes the underlying issue.
+	// Idempotency: a pass for this commit is reusable while it is still running
```
neversettle17-101, line 189: Defensive branch note (non-blocking): this path (`ok && existing.Status != Failed` after the running/verdicted reuse check) only fires for a non-running, non-failed, *empty-verdict* run. Since `UpdateReviewRunResult` only moves a run to `complete` with a `Valid()` verdict, that state can't arise through the normal API, so the "interrupted without ever calling Submit" comment describes an unreachable case. The guard is harmless, but the comment overstates how it's reached.

neversettle17-101 (commented):
**Verdict: Approve (ready).**

Re-review of the three new commits since 900efa5 (trigger idempotency gating + failing stale running passes, #342). The new logic is correct and well covered:

- `Trigger` now reuses a same-commit pass only while it is **running** or already **carries a verdict**; an interrupted non-running/unverdicted pass is superseded and a fresh pass starts, with a race-safe re-read (`GetReviewRun`) when `SupersedeReviewRun` no-ops because a verdict landed concurrently.
- `SupersedeStaleRunningReviewRuns` fails older running/unverdicted passes for *other* SHAs before launching the new commit's review, so a worker can't accumulate zombie running passes across commits. Correctly scoped (`target_sha != ?`, `status='running'`, `verdict=''`) and runs only on the create path, after the early-return reuse branches.
- sqlc regen matches the new queries; both supersede UPDATEs are guarded so they never clobber a verdicted or already-failed row. Tests exercise reuse-running, supersede-non-running, and supersede-older-commit. Build + `go test ./internal/review/... ./internal/service/review/... ./internal/lifecycle/...` pass locally.

Two minor notes (non-blocking):
1. The middle supersede branch and its comment cover a *non-running, non-failed, empty-verdict* run. Given `UpdateReviewRunResult` only completes with a `Valid()` verdict, that state isn't reachable through the normal API — the branch is purely defensive, so the comment ("interrupted without ever calling Submit") describes a path the state machine can't actually produce.
2. Prior-review nits are still open and remain non-blocking: `review.Engine.Submit` is dead in production (service reimplements it), and the `if !sent` branch in `ApplyReviewResult` (reactions.go:178) returns the same value as the line after it.

neversettle17-101 (commented):
**Verdict: Request changes — this commit does not compile.**

The only change in head commit `2f52055` ("Update review.go") edits the `Store` interface signature for `SupersedeReviewRun` from `ctx context.Context` to `ctx context`. `context` is the imported package, not a type, so this is a hard build break:

```
internal/review/review.go:36:25: context (package name) is not a type
```

`go build ./internal/review/...` fails at this commit. This looks like an accidental edit (the previous commit `675c38e` had the correct `ctx context.Context`). Please restore `ctx context.Context`:

```go
SupersedeReviewRun(ctx context.Context, id, body string) (bool, error)
```

Everything else from the prior review still stands as approved — this is the sole blocker.

backend/internal/review/review.go
```diff
@@ -34,6 +33,8 @@ type Store interface {
 	GetReviewBySession(ctx context.Context, id domain.SessionID) (domain.Review, bool, error)
 	InsertReviewRun(ctx context.Context, r domain.ReviewRun) error
 	UpdateReviewRunResult(ctx context.Context, id string, status domain.ReviewRunStatus, verdict domain.ReviewVerdict, body, githubReviewID string) (bool, error)
+	SupersedeReviewRun(ctx context, id, body string) (bool, error)
```
neversettle17-101: Build break: `context` is the imported package, not a type. The parameter must be `ctx context.Context`. Compiler error: `context (package name) is not a type`. Revert this back to `SupersedeReviewRun(ctx context.Context, id, body string) (bool, error)`.

neversettle17-101 (commented):
**Verdict: Approve (ready).**

The three new commits resolve every finding from the prior reviews, and the build break is gone — `go build ./...` and `go test ./internal/review/... ./internal/service/review/... ./internal/lifecycle/...` both pass at this commit (verified in a detached worktree).

- **Build break fixed** (`a9a0353`): the `context`-as-type typo is gone; the engine now imports `stdctx "context"` and uses it consistently across `review.go`. Compiles cleanly.
- **`engine.Submit` removed** (`f2c1e4d`): submit persistence is now single-sourced in `service.Submit`, eliminating the duplicated/diverging state machine I flagged earlier. The complete-branch retry path no longer silently mutates in-memory fields — it now rejects a retry that carries a *different* body or GitHub review id (`ErrInvalid`), with table tests covering both cases.
- **Dead branch removed** (`f2c1e4d`): the redundant `if !sent { return ReviewDeliverySent }` in `ApplyReviewResult` is gone.
- **Defensive-branch comment clarified** (`b9a9493`): the Trigger idempotency comment now accurately describes the fallback as defensive rather than a reachable interrupted-Submit path.

One purely cosmetic, non-blocking note: the `stdctx` alias for the standard `context` package is unusual (`context.Context` works without aliasing, and `service/review/review.go` still uses the plain import), so the two files are now stylistically inconsistent. Harmless — no change required.

neversettle17-101:
Screenshot 2026-06-22 at 10 46 40 AM
Tested the changes E2E locally. Merging the PR to unblock next features. Would take up any retroactive review comments.

PR #410, opened 2026-06-25
fix(frontend): split pending board idle state

laxmanclo:
## Summary
- rename the board's first work column to Pending
- split that column into a simple Working section and an Idle section
- collapse Idle down to a quiet count row when there are no idle sessions
- add a focused board test for the working/idle grouping

## Testing
- npm --prefix frontend test -- src/renderer/types/workspace.test.ts src/renderer/components/SessionsBoard.test.tsx

## Note
- Full frontend typecheck still hits the existing upstream Forge typing issue at forge.config.ts:33 (`osxNotarize`).


laxmanclo:
Closing because this change belongs in AgentWrapper/agent-orchestrator, not this repo.

PR #413, opened 2026-06-25
fix(frontend): require node 20 for electron dev

### LINKED ISSUE #412: Frontend install on Node 24/26 can leave Electron half-installed and break npm run dev
## Bug

Installing the Electron frontend on Node 24/26 can leave `node_modules/electron`
half-installed, which then makes `npm run dev` fail with:

```text
Error: Electron failed to install correctly, please delete node_modules/electron and try installing again
```

**Analyzed against:** `7c7c985` | **Confidence:** High

## Reproduction

```bash
cd frontend
node -v
# v26.2.0

npm install
npm run dev
```

Observed result:

```text
An unhandled rejection has occurred inside Forge:
Error: Electron failed to install correctly, please delete node_modules/electron and try installing again
```

Switching to Node 20 makes the same flow work:

```bash
cd frontend
nvm use 20 || nvm install 20
rm -rf node_modules
npm ci
npm run dev
```

## Root Cause

This is not a project-specific runtime bug in the app code. The frontend uses
Electron Forge + Electron, and newer Node 24/26 runtimes currently hit an
upstream Electron install failure that can leave `node_modules/electron`
partially extracted. In that state, Electron's expected install metadata is
missing and Forge later fails while locating the Electron executable.

ReverbCode's own desktop/frontend workflows already run on Node 20 today, so
contributors can hit this locally even though CI is green.

## Fix

Make the frontend's supported Node version explicit and fail fast before a
partial install:

- declare `engines.node = 20.x` in `frontend/package.json`
- add a `preinstall` guard that rejects unsupported Node majors with recovery steps
- add `frontend/.nvmrc`
- document the Node 20 requirement in the root README

## Impact

This is likely to affect contributors who use current Node 24/26 by default and
try the Electron frontend locally. The failure is confusing because `npm install`
can look mostly successful and the actual break only appears later on
`npm run dev`. A fast-fail guard removes that trap and aligns local setup with
the repo's existing Node 20 desktop workflow.

anirudh5harma:
Closes #412

## Problem

The Electron frontend is effectively Node-20-only today, but contributors are
not told that up front. On Node 24/26, upstream Electron/Electron Forge install
failures can leave `node_modules/electron` half-installed, and the break only
surfaces later when `npm run dev` fails with `Electron failed to install correctly`.

Our desktop/frontend workflows already run on Node 20, so this is local-dev
friction caused by missing guardrails rather than an app-specific code path.

## Change

- add `engines.node = 20.x` to `frontend/package.json`
- add a `preinstall` check that fails fast on unsupported Node versions with
  clear remediation steps
- add `frontend/.nvmrc`
- document the Node 20 requirement and the reason for it in the root README

## Test

```bash
cd frontend
export NVM_DIR="$HOME/.nvm"
. "$NVM_DIR/nvm.sh"

nvm use 20
npm test
```

Result:

```text
30 passed (30)
303 passed (303)
```

Unsupported-node guard:

```bash
node frontend/scripts/check-node-version.mjs
```

Verified:

- Node 26 fails immediately with the new explanatory message
- Node 20 passes the guard
- a fresh install on Node 26 now fails before a partial Electron install can
  corrupt `node_modules`

## Notes

`npm run typecheck` still reports the pre-existing `forge.config.ts` typing
error around `osxNotarize`; this change does not modify that area.


anirudh5harma:
Ported to AgentWrapper/agent-orchestrator#2284 and linked to AgentWrapper/agent-orchestrator#2206. Closing this legacy PR so review continues in the canonical repository.

PR #415, opened 2026-06-25
fix(frontend): canonicalize daemon identity paths

### LINKED ISSUE #414: Electron app shows empty board when daemon checkout path casing differs
## Bug

The Electron app can render an empty board even when the daemon has projects and
sessions, if the daemon reports the same checkout path with different casing.

Observed locally on macOS:

```text
daemon workingDirectory: /Users/anirudhsharma/documents/projects/reverbcode/backend
Electron expected cwd:  /Users/anirudhsharma/Documents/projects/ReverbCode/backend
```

Those paths resolve to the same directory on the local filesystem, but the app
treated them as different checkouts.

**Analyzed against:** `a4e0458` | **Confidence:** High

## Reproduction

On a macOS checkout where the daemon was started from a differently cased path:

```bash
cd frontend
nvm use 20
npm run dev
```

The daemon API still has data:

```bash
/tmp/ao project ls --json
/tmp/ao session ls --json
```

But the Electron renderer shows:

```text
No projects yet.
Click + above to register a git repo.
```

Inspecting the live Electron renderer showed:

```json
{
  "state": "error",
  "message": "Another AO daemon is already running from /Users/.../documents/projects/reverbcode/backend; expected this checkout at /Users/.../Documents/projects/ReverbCode/backend."
}
```

At the same time, direct renderer fetches to `/api/v1/projects` and
`/api/v1/sessions` returned the expected project and sessions.

## Root Cause

`frontend/src/main.ts` compared daemon identity paths with string-normalized
`path.resolve()` values. That only lowercased paths on Windows.

macOS filesystems are often case-insensitive but case-preserving, so the same
directory can be observed as both `documents/projects/reverbcode` and
`Documents/projects/ReverbCode`. The daemon identity check then returned an
error status, and the renderer refused to trust the daemon API base URL.

## Fix

Canonicalize path identity through `realpathSync.native()` before comparing
daemon checkout paths. Preserve the existing behavior of rejecting genuinely
different checkouts, and keep Windows comparisons case-insensitive.

The fix also handles non-existent child paths by canonicalizing the nearest
existing ancestor before applying the remaining path segments.

## Impact

This can make the Electron app look empty or unable to use existing daemon data
even though the daemon is healthy and API calls return projects/sessions. It is
especially likely on macOS when a checkout or daemon was launched through paths
with different casing.

anirudh5harma:
Closes #414

## Problem

The Electron supervisor can reject a healthy daemon from the same checkout when
macOS reports the checkout path with different casing. In that state the
renderer shows the empty project board even though `/api/v1/projects` and
`/api/v1/sessions` return real data.

## Change

- move daemon path identity checks into a shared `path-identity` helper
- canonicalize paths with `realpathSync.native()` before comparing them
- preserve Windows case-insensitive comparisons
- canonicalize the nearest existing ancestor for not-yet-created child paths
- cover same-path and child-path behavior with focused tests

## Tests

```bash
cd frontend
export NVM_DIR="$HOME/.nvm"
. "$NVM_DIR/nvm.sh"
nvm use 20

npm test -- --run src/shared/path-identity.test.ts src/shared/daemon-attach.test.ts
npm test
```

Results:

```text
2 passed (2), 34 passed (34)
31 passed (31), 307 passed (307)
```

## Verification

Verified against the live Electron renderer through the Chromium DevTools
protocol after restarting `npm run dev`:

- `window.ao.daemon.getStatus()` reports `state: "ready"`
- the board renders `demo-repo`
- the board renders worker sessions `demo-repo-1` and `demo-repo-2`

### DIFF
--- a/frontend/src/main.ts
+++ b/frontend/src/main.ts
@@ -33,6 +33,7 @@ import { buildDaemonEnv, resolveShellEnv, type ShellRunner } from "./shared/shel
 import { DEFAULT_POSTHOG_HOST, DEFAULT_POSTHOG_PROJECT_KEY } from "./shared/posthog-config";
 import { buildTelemetryBootstrap } from "./shared/telemetry";
 import { createBrowserViewHost, type BrowserViewHost } from "./main/browser-view-host";
+import { pathInside, samePath } from "./shared/path-identity";
 
 // Globals injected at compile time by @electron-forge/plugin-vite.
 declare const MAIN_WINDOW_VITE_DEV_SERVER_URL: string | undefined;
@@ -305,21 +306,6 @@ function daemonEnv(): NodeJS.ProcessEnv {
 		return { ...process.env, ...telemetryOverrides() };
 	}
 	return buildDaemonEnv(process.env, cachedShellEnv, telemetryOverrides());
-}
-
-function pathKey(value: string): string {
-	const resolved = path.resolve(value);
-	return process.platform === "win32" ? resolved.toLowerCase() : resolved;
-}
-
-function samePath(a: string, b: string): boolean {
-	return pathKey(a) === pathKey(b);
-}
-
-function pathInside(child: string, parent: string): boolean {
-	const childKey = pathKey(child);
-	const parentKey = pathKey(parent);
-	return childKey === parentKey || childKey.startsWith(parentKey + path.sep);
 }
 
 function processAlive(pid: number): boolean {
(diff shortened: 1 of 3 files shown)


anirudh5harma:
Ported to AgentWrapper/agent-orchestrator#2285, which closes AgentWrapper/agent-orchestrator#2279. Closing this legacy PR so review continues in the canonical repository.

PR #419, opened 2026-06-26
fix(frontend): restore verification commands

### LINKED ISSUE #418: Frontend verification commands fail on clean main
## Bug

The documented frontend verification commands do not work cleanly on `main`.

## Reproduction

From a clean checkout:

```bash
cd frontend
npm run typecheck
```

Observed result:

```text
forge.config.ts(33,51): error TS2339: Property 'osxNotarize' does not exist on type 'ForgePackagerOptions | undefined'.
```

The repo guidance also tells contributors to run:

```bash
cd frontend
npm run build
```

but `frontend/package.json` does not define a `build` script.

## Root Cause

`forge.config.ts` casts the keychain-profile notarization branch through `ForgeConfig["packagerConfig"]["osxNotarize"]`. Because `packagerConfig` is optional on the public Forge config type, TypeScript rejects that indexed access.

The frontend workflow currently skips typecheck because of this pre-existing type error, so frontend PRs can pass CI while TypeScript regressions remain hidden.

## Fix

Use the installed notarization credential type directly for the keychain-profile branch, add a real frontend `build` script, and restore `npm run typecheck` in the frontend CI workflow.

## Impact

Contributors can run the documented frontend checks locally, and CI regains the TypeScript safety net for frontend changes.

anirudh5harma:
## Summary

Closes #418.

Frontend verification works again on clean checkouts: `npm run typecheck` no longer fails on the Forge notarization config, and `npm run build` now exists as the documented package-build command.

The frontend workflow now runs typecheck before vitest, restoring the TypeScript guard that was previously disabled because of the pre-existing config error.

## Tests

- `cd frontend && npm run typecheck`
- `cd frontend && npm run build`
- `cd frontend && npm test -- --run`
- `git diff --check`

### DIFF
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -5,10 +5,6 @@ name: Frontend
 # vitest.config.ts, and the repo had neither until #171) — this job is the
 # guard against that happening again.
 #
-# Typecheck is intentionally NOT run here yet: forge.config.ts and
-# update-electron-app carry pre-existing type errors. Add `npm run typecheck`
-# once those are fixed.
-
 on:
   push:
     branches: [main]
@@ -36,6 +32,9 @@ jobs:
           cache-dependency-path: frontend/package-lock.json
 
       - run: npm ci
+
+      - name: Typecheck
+        run: npm run typecheck
 
       - name: Run vitest suite
         run: npx vitest run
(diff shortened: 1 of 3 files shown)


anirudh5harma:
Ported and updated for current main as AgentWrapper/agent-orchestrator#2287, closing AgentWrapper/agent-orchestrator#2281. Closing this legacy PR.

